PR #1, opened 2017-09-20
Travis syntax check and lint

manics:
Adds syntax checking and ansible-lint.

This is expected to fail on travis with the current playbooks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+sudo: required
+language: python
+
+services:
+  - docker
+
+install:
+  - pip install ome-ansible-molecule-dependencies
+
+script:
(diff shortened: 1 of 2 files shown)


sbesson:
Activated Travis. Closing/reopening to validate this PR.

sbesson:
Some syntax issues (a lot of trailing whitespaces). What is the next action here? Should we cleanup existing playbooks before adding anything enw?

manics:
It's a matrix build, so new playbooks can be tested and verified to pass without needing to fix existing ones immediately.

sbesson:
That's fine, should we mark the existing ones as expected to fail in that case and we'll review them as we modify them?

manics:
updated

PR #2, opened 2017-09-20
Fix website playbook lint errors

sbesson:
Following #1, this PR adds the website playbook `www.yml`  to the Travis matrix for verification.

Although we might switch to playbook directories with their own `molecule.yml`, this PR makes minimal modification to the existing `.travis.yml` logic:
- the Travis execution is now performed under the subdirectory if applicable including the copying of `molecule-syntaxcheck.yml`
- the `www/www.yml` lint errors are fixed including trailing whitespaces and the usage of `state: installed` for the package installation

To test this PR, check that Travis is green including the check of the `www` playbook and that the changes diff is sensible.


manics:
I don't think it's essential to keep the current `.travis.yml` layout. In fact it might be easier to put playbooks in their own directory so (eventually) you could have individual molecule.yml files for each.

kennethgillen:
@sbesson - feel free to change all the open-vm-tools invocations to `state: present` or `state: installed` - it's safer.

www/www.yml
```diff
@@ -8,13 +8,13 @@
...
         name: open-vm-tools
-        state: latest
+        state: installed
       when: >
            ((ansible_virtualization_type is defined)
            and (ansible_virtualization_type == "VMware"))
 
-    # Two tasks here which could instead ues the role at 
-    # https://github.com/openmicroscopy/ansible-role-lvm-partition/, but that role 
+    # Two tasks here which could instead ues the role at
```
mtbc: use

joshmoore:
Happy with these minimal changes and the green status. Follow-ups are welcome to refine the directory structure. Merging.

NB: as far as I know, a tagging policy for this repo hasn't been discussed.

PR #5, opened 2017-10-05
Move jekyll into a separate playbook

manics:
Follow-up to a brief discussion on docs deployment:

This moves the jekyll docs deployment into a separate playbook. This should make it easier in future to give limited sudo permissions for an automated user to update the docs without requiring full access to the server to run the whole www.yml playbook.

Also refactors the `www.yml` playbook so that steps which aren't dependent on hardware or LSC can be tested with molecule.yml on travis.

Note that this playbook is NOT idempotent:
- jekyll build is forcibly rebuilt (this is expected)
- Both `NGINX - Performance tuning` steps conflict with the configuration file managed by the nginx-proxy role.

----
A future refinement could be to refactor the jekyll-build role to optionally run without `sudo`, so the docs could be built be a dedicated non-admin user.

### DIFF
--- a/www/www-jekyll.yml
+++ b/www/www-jekyll.yml
@@ -1,0 +1,14 @@
+# Install NGINX, and prepare the OME (UoD/SLS) prerequisites
+
+- hosts: www
+  environment:
+      PATH: /usr/local/bin:{{ ansible_env.PATH }}
+
+  roles:
+
+    - role: openmicroscopy.jekyll-build
+      tags: jekyll
+      jekyll_build_git_repo: "https://github.com/openmicroscopy/www.openmicroscopy.org"
+      jekyll_build_force_rebuild: True
+      jekyll_build_config: ['_config.yml', '_prod.yml']
+      jekyll_build_name: "{{ website_name }}"
--- a/www/www.yml
+++ b/www/www.yml
@@ -79,13 +79,6 @@
       tags: monitoring
       when: "'10.1.255.216' in ansible_dns.nameservers"
 
-    - role: openmicroscopy.jekyll-build
-      tags: jekyll
-      jekyll_build_git_repo: "https://github.com/openmicroscopy/www.openmicroscopy.org"
-      jekyll_build_force_rebuild: True
-      jekyll_build_config: ['_config.yml', '_prod.yml']
-      jekyll_build_name: "{{ website_name }}"
-
     - role: openmicroscopy.nginx_proxy
       tags: nginxconf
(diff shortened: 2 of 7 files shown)


sbesson:
@kennethgillen did you test this refactoring as part of the website redeployment?

www/www.yml
```diff
@@ -148,107 +93,6 @@
...
-    - match: "~/site/support/?$"
-      dest: /docs
-    - match: "~/site/support/(?<link>.*)$"
-      dest: https://www-legacy.openmicroscopy.org/site/support/$link
 
-    nginx_proxy_direct_locations:
-    - location: "/"
-      root: "/var/www/{{ website_name }}/html"
-      index: index.html
+- include: www-deployment.yml
```
kennethgillen: The file was called `www-deploy.yml`, no?

kennethgillen:
@sbesson  - no sir, I think it came afterwards.

www/www-jekyll.yml
```diff
@@ -0,0 +1,14 @@
+# Install NGINX, and prepare the OME (UoD/SLS) prerequisites
+
+- hosts: www
+  environment:
+      PATH: /usr/local/bin:{{ ansible_env.PATH }}
+
+  roles:
+
+    - role: openmicroscopy.jekyll-build
+      tags: jekyll
```
sbesson: Does the split of the main playbooks into sub-playbooks makes the usage of tags redundant e.g as you should be able to run `www-jekyll.yml` in isolation?

manics: Can of worms 😊. Only thing I'd say is if we want to use tags in other playbooks we need to standardise them across the OME, otherwise we'll have problems when trying to use them with something like site.yml.

sbesson:
Tested the lightweight deployment of the website with 

```
$ ansible-playbook --ask-become playbooks/www/www-jekyll.yml
```

All worked as expected and the OMERO 5.4.0 downloads/announcements were updated without requiring a run of the complete playbook.

Also tested in dry-run mode the execution of the full playbook:

```
$ ansible-playbook --ask-become playbooks/www/www.yml
```

So in conclusion, the current workflow is preserved but it is possible to execute sub-playbooks individually depending on the needs. Happy for this to go in if @kennethgillen has no objection.

sbesson:
Having read no veto for the last week and since this change is modifying `.travis.yml` again, I am merging so that we can work through the reinclusion of the demo/nightshade playbooks as the next step.

PR #10, opened 2017-11-23
Add routine for installing Figure_To_Pdf.py script

pwalczysko:
Encouraged by @manics , I am adding the routine to for installing the Figure export script to the nightshade playbook (first commit) and the demo server playbook (second commit). The two added paragraphs are working fine already on the "outreach" server playbook, see https://github.com/pwalczysko/ansible-examples-omero/commit/5ece2affaed25c0432a4fd77ddd8e0aa37b040ba. 

To test, see that after you ran the playbooks, the figure pdf creation is enabled and works as expected.

cc @joshmoore @kennethgillen @manics

### DIFF
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -300,17 +300,16 @@
         state: directory
         mode: 0755
         recurse: yes
-        owner: "omero-server"
-        group: "omero-server"
+        owner: root
 
     - name: Download the Figure_To_Pdf.py script
       become: yes
       get_url:
         url: https://raw.githubusercontent.com/ome/omero-figure/v3.1.2/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
         dest: /opt/omero/server/OMERO.server/lib/scripts/omero/figure_scripts/Figure_To_Pdf.py
-        mode: 0755
-        owner: "omero-server"
-        group: "omero-server"
+        mode: 0644
+        owner: root
+        force: yes
 
   vars:
     # Check_MK (system monitoring) paths
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -293,6 +293,25 @@
         dest: /etc/cron.daily/nightly-pg_dump-omero.sh
         mode: "u=rwx,go="
 
+    - name: Create a figure scripts directory
+      become: yes
+      file:
+        path: /opt/omero/server/OMERO.server/lib/scripts/omero/figure_scripts
+        state: directory
+        mode: 0755
+        recurse: yes
+        owner: "omero-server"
+        group: "omero-server"
+
+    - name: Download the Figure_To_Pdf.py script
+      become: yes
+      get_url:
+        url: https://raw.githubusercontent.com/ome/omero-figure/v3.1.2/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+        dest: /opt/omero/server/OMERO.server/lib/scripts/omero/figure_scripts/Figure_To_Pdf.py
+        mode: 0755
+        owner: "omero-server"
+        group: "omero-server"
+
   vars:
     # Check_MK (system monitoring) paths
     check_mk_agent_plugin_path: /usr/share/check-mk-agent/available-plugins
--- a/ome-dundeeomero.yml
+++ b/ome-dundeeomero.yml
@@ -228,6 +228,25 @@
         dest: /etc/cron.daily/nightly-pg_dump-omero.sh
         mode: "u=rwx,go="
 
+    - name: Create a figure scripts directory
+      become: yes
+      file:
+        path: /opt/omero/server/OMERO.server/lib/scripts/omero/figure_scripts
+        state: directory
+        mode: 0755
+        recurse: yes
+        owner: "omero-server"
+        group: "omero-server"
+
+    - name: Download the Figure_To_Pdf.py script
+      become: yes
+      get_url:
+        url: https://raw.githubusercontent.com/ome/omero-figure/v3.1.2/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
+        dest: /opt/omero/server/OMERO.server/lib/scripts/omero/figure_scripts/Figure_To_Pdf.py
+        mode: 0755
+        owner: "omero-server"
+        group: "omero-server"
+
   vars:
     # Check_MK (system monitoring) paths
     check_mk_agent_plugin_path: /usr/share/check-mk-agent/available-plugins
--- a/ome-dundeeomero.yml
+++ b/ome-dundeeomero.yml
@@ -243,7 +243,7 @@
       get_url:
         url: https://raw.githubusercontent.com/ome/omero-figure/v3.1.2/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
         dest: /opt/omero/server/OMERO.server/lib/scripts/omero/figure_scripts/Figure_To_Pdf.py
-        mode: 0755
+        mode: 0400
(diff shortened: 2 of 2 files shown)


ome-dundeeomero.yml
```diff
@@ -228,6 +228,25 @@
...
         mode: "u=rwx,go="
 
+    - name: Create a figure scripts directory
+      become: yes
+      file:
+        path: /opt/omero/server/OMERO.server/lib/scripts/omero/figure_scripts
+        state: directory
+        mode: 0755
+        recurse: yes
+        owner: "omero-server"
```
kennethgillen: @pwalczysko - Thanks for this PR! This has been on my wish list for a while. In this case, the omero server O.S. user is unique to this host - see line :172.

ome-demoserver.yml
```diff
@@ -293,6 +293,25 @@
...
+        state: directory
+        mode: 0755
+        recurse: yes
+        owner: "omero-server"
+        group: "omero-server"
+
+    - name: Download the Figure_To_Pdf.py script
+      become: yes
+      get_url:
+        url: https://raw.githubusercontent.com/ome/omero-figure/v3.1.2/omero_figure/scripts/omero/figure_scripts/Figure_To_Pdf.py
```
kennethgillen: We'll need to find out if it's safe to mix versions of the figures script with different versions of figure. This figure version "3.1.2" is not hard-coded anywhere else in the playbook, and you may be pulling in a newer version of the script than the accessing OMERO.web client.

@openmicroscopy/devops have also discussed hard-coding the versions of the OMERO.web plugins on each of the systems, so if we go ahead with that strategy, then hard-coding the version here is fine - we'd just need to change it alongside the version of figure. Or have this and the "install figure" statement parse a variable defined further up the playbook with the version defined.

ome-dundeeomero.yml
manics: Does anyone know whether the user actually matters here, or could it be owned by root?

ome-demoserver.yml
manics: @pwalczysko As discussed on one of your other playbooks, you'll need to either add `force: yes` or add a checksum

joshmoore: The server doesn't need to edit the location, only be able to read it.

ome-dundeeomero.yml
pwalczysko: @kennethgillen 
>omero server O.S user is unique....
Sorry, do not understand the context of this remark. What does the uniquenness imply ? What should I do about it ?

joshmoore: @pwalczysko meaning that the user that owns OMERO differs across hosts. Nightshade uses a different name.

pwalczysko: @joshmoore : Thank you. And the name is.. ? The line 172 mentioned by @kennethgillen resolves to me as https://github.com/openmicroscopy/prod-playbooks/pull/10/files#diff-896819b5bdf43f197f838c7cc66babf4R172 - I know the context now, but not the answer :)

kennethgillen: @pwalczysko - line 172 shows an example of existing usage of that user within the context of this playbook.

kennethgillen: Mentioned to @pwalczysko was referring to `ome-dundeeomero.yml` rather than `ome-demoserver.yml`, and that my line :172 was a few lines out - was referring to :169 https://trello.com/c/1qSXnrPt/13-demo-541-upgrade#comment-5a266dcb5d638979135e3fe3

ome-demoserver.yml
pwalczysko: @joshmoore : so I should change the line https://github.com/openmicroscopy/prod-playbooks/pull/10/files#diff-896819b5bdf43f197f838c7cc66babf4R311 to 0500 (read and execute by owner), right ?

kennethgillen: It's not a script in the "linux/execute" sense, so won't need `+x`. And to allow replacement and deletion via CLI, the server might want to be able to write it too, so `0600`?

manics: If you're managing this with Ansible the server shouldn't need write access.

kennethgillen: That's true, provided we don't test it on production then. As it's extra hassle to (remember to) change the perms during a broken production deploy, when I'd expect `omero upload` to work.

manics: If production breaks doesn't that imply the playbook is broken and should be fixed?

kennethgillen: Yes, but coming to that conclusion doesn't help a live out-of-hours deployment. We can go with your suggestion, and hopefully it's tested on a non-prod system then.

manics: Ideally this would be refactored so it's testable by travis.

manics: If there's a deeper reason you see the figure script as being prone to failures can you add that to a card and we should try and fix it? The whole playbook should be equally trusted for a production server.

jburel: first test to add is for the script itself so we can test it via travis
the infra-work will help on that front

kennethgillen: I've just never deployed it like this myself, so don't know what to expect. And I see a mechanism for "escape/fixing" as a sysadmin as being removed by the changing of the permissions.

sbesson:
Note with https://github.com/openmicroscopy/prod-playbooks/pull/26 and the training server moving to the `prod-playbooks` repository, we might consider refactoring this logic e.g. in a subplaybook included by all playbooks (training, omedundee-omero, omedemo).

pwalczysko:
https://github.com/openmicroscopy/prod-playbooks/pull/10/commits/307794023708f573121ffb433a3043ad90c666e3 reflects the results of discussion with @manics in the office. His suggestion was to have "root" as the owner of the directory and script, which should satisfy @joshmoore 's and @kennethgillen 's comments in that regard.

pwalczysko:
All comments should be addressed now except
    - https://github.com/openmicroscopy/prod-playbooks/pull/10#discussion_r154902720 which @kennethgillen explicitly encouraged me to do in another PR
    - https://github.com/openmicroscopy/prod-playbooks/pull/10#issuecomment-352794656 -> @sbesson 's comment seems to be for another PR too (?)

sbesson:
Re https://github.com/openmicroscopy/prod-playbooks/pull/10#issuecomment-352794656, I think the main remaining question is whether there we can avoid the copy-n-paste of the same block of tasks across 2 (or more) separate playbooks.

https://github.com/IDR/deployment/ makes extensive use of `include` tasks although I see that Ansible 2.4 is evolving this concept (http://docs.ansible.com/ansible/2.4/playbooks_reuse.html). @manics might have more hindsight on the impact of using this functionality.

manics:
Main points about the structure in https://github.com/IDR/deployment/
- single layer of includes
- clear separation between parent playbooks (`idr-NN-*`) which only include others, and the rest
- parent playbooks are intended to be run standalone after the initial full deployment, whereas other playbooks may or may not leave the system in the expected state when run on their own.

joshmoore:
:+1: for Simon's delineation of _types_ of playbooks (top, include only, mixed?). NB: I've always had issues remembering what's what though, so somewhere (not on this PR) I'd suggest we gather all the needed types and have YANC (yet another naming conversation).

joshmoore:
Few follow-on cards::

 * https://trello.com/c/dHgWG352/4475-ansible-playbook-types
 * https://trello.com/c/A7CEt8JY/4476-ansible-use-of-root

Otherwise, :+1: thanks, @pwalczysko. This will be run against nightshade for the 5.4.2 release.

PR #7, opened 2017-10-13
Nightshade to 540

kennethgillen:
This PR builds on top of https://github.com/openmicroscopy/prod-playbooks/pull/4 - Not sure if this branch on the other branch is the right way to go about this. 

Merged in the .travis.yml from master to avoid conflicts, thanks to @jburel for the tip.

PR #11, opened 2017-11-28
Demo server: Bump versions for 540 upgrade.

kennethgillen:
OMERO and Web versions for demo sever prod playbook incremented for demo server upgrade.

### DIFF
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -120,7 +120,7 @@
         databases: ["{{ vault.omero_server_dbname }}"]
 
     - role: openmicroscopy.omero-server
-      omero_server_release: 5.3.4
+      omero_server_release: 5.4.0
       omero_server_dbuser: "{{ vault.omero_server_db_user }}"
       omero_server_dbname:  "{{ vault.omero_server_dbname }}"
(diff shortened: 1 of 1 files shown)


kennethgillen:
I'm not sure what, if anything, Travis is doing here. Perhaps we can sort that out later? Travis didn't go well on https://github.com/openmicroscopy/prod-playbooks/pull/7/commits

joshmoore:
Looks green to me (syntax check etc). Are you pushing the 5.4.1 changes to this PR as well or do you want to get it in as is?

kennethgillen:
I'd be inclined to get in as the recorded state of the 5.4.0 upgrade. Then potentially getting more Travis etc into the mix for 5.4.1.

joshmoore:
WFM :+1:

sbesson:
Re https://github.com/openmicroscopy/prod-playbooks/pull/11#issuecomment-347577624, @josh said it all. As the syntax check in `demo` seems to be green, probably worth removing it from the allowed failures in https://github.com/openmicroscopy/prod-playbooks/blob/master/.travis.yml#L38 now too.

PR #12, opened 2017-11-28
Remove ome-demoserver playbook from the allowed failures list

sbesson:
See https://github.com/openmicroscopy/prod-playbooks/pull/11#issuecomment-347582513

This should increase the strictness of the demo playbook by failing Travis if the syntax is invalid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,5 @@ env:
 matrix:
   allow_failures:
   - env: DIRECTORY="." PLAYBOOK=nightshade-web.yml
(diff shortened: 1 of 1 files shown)


kennethgillen:
👍  - looks good. I have to learn how to interpret travis output for the 541 upgrades PRs - I didn't think it was doing anything.

sbesson:
Thanks @kennethgillen. Merging and the Travis output parsing can be discussed if the builds turn red.

PR #13, opened 2017-11-29
OME Production Server OMERO upgrades (540-541)

kennethgillen:
OME Production Server OMERO upgrades (540-541)

### DIFF
--- a/nightshade-webclients.yml
+++ b/nightshade-webclients.yml
@@ -43,15 +43,28 @@
       pip:
         name:
         - "omero-figure"
-        - "omero-fpbioimage"
         - "omero-webtagging-autotag"
         - "omero-webtagging-tagsearch"
+        editable: False
+        state: "{{ webapps_state }}"
+        # variable comes from role openmicroscopy.omero-web
+        virtualenv: "{{ omero_web_basedir }}/venv"
+        virtualenv_site_packages: yes
+      notify:
+        - restart omero-web
+
+    - name: Omero.web plugins | plugin install via pip & pypi
+      become: yes
+      pip:
+        name:
+        - "omero-fpbioimage"
         - "omero-iviewer"
         editable: False
         state: "{{ webapps_state }}"
         # variable comes from role openmicroscopy.omero-web
         virtualenv: "{{ omero_web_basedir }}/venv"
         virtualenv_site_packages: yes
+      when: "'ns-web.openmicroscopy.org' in inventory_hostname"
       notify:
         - restart omero-web
 
@@ -138,6 +151,20 @@
         owner: "root"
         group: "root"
         mode: "u=rw,go=r"
+      notify:
+        - restart omero-web
+
+    # Config for OMERO.web plugins, loaded into OMERO.web by the
+    # omero.web systemd restart.
+    - name:
+      become: yes
+      template:
+        src: templates/omero-web-config-for-webapps_ns-web-pub.j2
+        dest: "{{ omero_web_basedir }}/config/omero-web-config-for-webapps.omero"
+        owner: "root"
+        group: "root"
+        mode: "u=rw,go=r"
+      when: "'ns-web-pub.openmicroscopy.org' in inventory_hostname"
       notify:
         - restart omero-web
 
--- a/nightshade-webclients.yml
+++ b/nightshade-webclients.yml
@@ -149,6 +149,7 @@
         owner: "root"
         group: "root"
         mode: "u=rw,go=r"
+      when: "'ns-web.openmicroscopy.org' in inventory_hostname"
       notify:
         - restart omero-web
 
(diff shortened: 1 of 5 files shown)


sbesson:
As we are planning server downtime, a couple of suggestions/points of discussion for now or later:

- bump `omero-marshal` to 0.5.2
- specify the version all OMERO.webapps i.e. `omero-figure==3.1.2` instead of `omero-figure`
- bump the version of the server/postgresql roles

sbesson:
Briefly discussed with @kennethgillen who pointed me at the current `upgrade_webapps` strategy. Trying to summarize the  outcome of the discussion about the pros and cons managing the Web apps versions via the playbook/variables:

Cons:
- maintenance: puts the burden on the playbook maintainer to look for which app has been released, the version changes...

Pros:
- representation: the playbook gives a representation of the complete state of the Web including all web apps
- communication: PRs express the components that will and will not be upgraded. NB: in its initial state, I could not tell if the planned upgrade would include the web apps or not as this is a runtime argument

Happy to hear others' thoughts on this question. If we were to switch to a versioned strategy, it might be that we want to have  a more proactive behavior following the component release where we immediately the new version of the component as a PR against our prod systems.

manics:
Assuming the web app version are all in this repo we could put the onus on whoever wishes to release a new web-app version to open a PR here after release (even if it's put on hold till)?

kennethgillen:
@openmicroscopy/devops : requesting review on last commit. 

I tried a second playbook with the host specifically `ns-web-pub` following this playbook in the same file, at the suggestion of @manics , which looked nicer, but I was unable to use the handler `restart nginx`, presumably because this second playbook doesn't use any roles. 

I couldn't find a way to import the role without running it, to have access to the handlers.

kennethgillen:
I don't think `--check` and `--diff` would fully test since I'm using `blockinfile` - as the play could continually add the same block to the text file, which wouldn't happen in `--check`.

Any @openmicroscopy/devops is welcome to test this with infra-testpr as follows:

```
(ansible)LS27172:ansible kenny$ echorun ansible-playbook  -l infra-testpr.openmicroscopy.org -i ~/forks/management_tools/ansible/inventory/test-hosts  ~/forks/prod-playbooks/nightshade-webclients.yml --extra-vars omero_web_upgrade=True --extra-vars upgrade_webapps=True --ask-become-pass  --tags "redirect"
```

manics:
> I couldn't find a way to import the role without running it, to have access to the handlers.

Quickest way: Add a handlers: section to this playbook and redefine restart nginx.
Longer: Create a new role containing only handlers that could be reused.

If you're staying with the current approach you might as well combine the two `blockinfile` tasks into one.

kennethgillen:
I don't know what's best at this point. More work and duplication redefining the handler? As it stands I can't combine the `blockinfile` tasks, as one is conditional on hostname, the other is not.

joshmoore:
Discussed with @manics & @kennethgillen : 

 * using this strategy (blockinfile) to release 5.4.1 to production on Monday
 * even though blockinfile should eventually be linted away: https://github.com/openmicroscopy/ome-ansible-molecule-dependencies/issues/9
 * longer-term, [nginx-location](https://github.com/openmicroscopy/ansible-role-omero-web/issues/8) should be used to make maintenance simpler
 * today we will try to get in other changes (disable iviewer, etc) to improve the user experience to the extent that we don't have to push the deployment again

cc: @jburel @sbesson

kennethgillen:
Note sister PR at https://github.com/openmicroscopy/management_tools/pull/603

kennethgillen:
@sbesson  - wrt omero-marshall - from my perspective  https://github.com/openmicroscopy/ansible-role-omero-web/commit/610e0af09d7bee49b959dd5f7f74c5f4443c0abe#diff-14acf9d7d5d537ebe69bddd620416675 fixes having to hard-code it at all, by installing the requirements from the downloaded OMERO dist. Does that match your expectation?

sbesson:
:+1: for using the requirements from the downloaded OMERO dist from my perspective as this matches the sysadmin documentation.

One small question with regard to the semantics: the `state:present` in https://github.com/openmicroscopy/ansible-role-omero-web/commit/610e0af09d7bee49b959dd5f7f74c5f4443c0abe#diff-14acf9d7d5d537ebe69bddd620416675 is slightly different from the `pip install -U` step in https://docs.openmicroscopy.org/omero/5.4.1/sysadmins/omeroweb-upgrade.html#dependencies which is rather `state: latest`?

kennethgillen:
Successful upgrade of ome-dundeeomero and ns-web* complete, once required role versions were worked out through trial-and-error. 

Follow up tomorrow am with Demoserver.

joshmoore:
Note: the state of this run excluded `omero-user.yml` and `omero-datadir.yml` after having reverted to an older version that didn't have a `when:` clause.

joshmoore:
Merging as the current state of nightshade in production. A follow-up PR will bump all the roles back to the latest and try to handle the issues mentioned above.

PR #14, opened 2017-11-30
Web proxy

joshmoore:
Currently testing with:

```
$ ansible-playbook -i ../ansible/inventory/prod-hosts \
    -l web-proxy* playbooks/web-proxy/playbook.yml \
    -C --diff -t config --ask-become
```

Note: this needs the requirements updates from gh-9. I won't push anything here until it's been merged.

### DIFF
--- a/web-proxy/playbook.yml
+++ b/web-proxy/playbook.yml
@@ -2,6 +2,7 @@
 # Playbook for maintaining IDR service nodes
 
 - hosts: prod-web-proxies
+  tags: setup
   roles:
   - role: network
   - role: openmicroscopy.lvm-partition
@@ -14,4 +15,8 @@
     lvm_lvmount: /var/log
     lvm_lvsize: "{{ varlog_size }}"
     lvm_lvfilesystem: "{{ root_filesystem }}"
+
+- hosts: prod-web-proxies
+  tags: config
+  roles:
   - role: openmicroscopy.nginx-proxy
--- a/web-proxy/playbook.yml
+++ b/web-proxy/playbook.yml
@@ -4,7 +4,7 @@
 - hosts: prod-web-proxies
   tags: setup
   roles:
-  - role: network
+  - role: openmicroscopy.network
   - role: openmicroscopy.lvm-partition
     lvm_lvname: root
     lvm_lvmount: /
--- a/web-proxy/playbook.yml
+++ b/web-proxy/playbook.yml
@@ -2,7 +2,6 @@
 # Playbook for maintaining IDR service nodes
 
 - hosts: prod-web-proxies
-  tags: setup
   roles:
   - role: openmicroscopy.network
   - role: openmicroscopy.lvm-partition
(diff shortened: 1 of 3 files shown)


web-proxy/playbook.yml
```diff
@@ -14,4 +15,8 @@
     lvm_lvmount: /var/log
     lvm_lvsize: "{{ varlog_size }}"
     lvm_lvfilesystem: "{{ root_filesystem }}"
+
+- hosts: prod-web-proxies
+  tags: config
```
manics: I'll need a formal definition of what the `config` tag means to us to fully evaluate this, but at first glance this is misleading since this also installs Nginx, it's not just configuration. If the intention is hardware/system-initialisation/low-level-setup tasks which aren't tied to a specific system maybe come up with a different tag name?

joshmoore: It's not a formal definition (and I'd argue a lot of our ansible setup doesn't have one :smile:) but in the spirit of what you're asking for:

For me, `config` meant "do the configuration of the proxy" or more generically, "do the actual work of this playbook (i.e. apply config)" rather than any of the preliminary "setup" tasks. I certainly agree that the way the playbook is currently configured, this tag **doesn't** achieve that but it certainly reduced the steps that were needed, and some that seemed to have changes I wasn't comfortable with.

If I was to think of a more complete list of phases I have a feeling for:

 * bootstrap (naming from prod-playbook; really only needed once)
 * install (of requirements; safe to only be called once
 * update (apply as needed; could also be something like 'state')
 * assert (checks state; apply as frequently as possible)

This should also be compared against your list from idr/deployment's deploy-idr.sh script.

manics: The approach taken in idr/deployment is to split playbooks first by provisioning which will generally by run once (mostly storage and networking config, plus an initial yum ugprade), deployment which is split by application and can be run as often as wanted (for example, nginx, omero, haproxy), then the kubernetes stuff which is still relatively new so doesn't yet fit in with the rest of the playbooks.

Tags are useful for manual deployment, especially for testing. However if you feel these playbooks aren't in a state where you're happy to run them in full at regular intervals (e.g. by Jenkins, Ansible Tower, cron) we should fix that, for instance by having clear comments/READMEs to indicate the state of everything, or by splitting up playbooks.

joshmoore: I've reverted the tags since this has now been applied in its entirety on production. When I was the first person re-running these playbooks after an extended time, I was *not* comfortable re-running this.  I also think that the nginx-proxy should do less or provide tag endpoints to reduce the impact of trying to `update` something (based on the semantics above).

joshmoore: Linked to this discuss from https://github.com/openmicroscopy/prod-playbooks/issues/24

sbesson:
No veto to getting this in assuming it fixes the deployment of `web-proxy`.
The changes in `www/requirements.yml` are somewhat unexpected and since the additions start to duplicate the roles in the top-level `requirements.yml`, I wonder if we should consider unifying these requirement files after this PR.

joshmoore:
:+1: for refactoring.

PR #17, opened 2017-12-05
ome demoserver to 5.4.1

kennethgillen:
Upgrading ome-demoserver to OMERO 5.4.1

(Sibling PR: https://github.com/openmicroscopy/management_tools/pull/610/)

### DIFF
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -126,9 +126,9 @@
       omero_server_dbpassword: "{{ vault.omero_server_dbpassword }}"
       omero_server_rootpassword: "{{ vault.omero_server_rootpassword }}"
       omero_server_systemd_limit_nofile: 16384
-      # Once omego version bug is fixed, we can remove this hard-coding.
+      # TODO: Once omego version bug is fixed, we can remove this hard-coding.
       # It should be defined in the omero-server role.
-      # See: https://github.com/openmicroscopy/ansible-role-omego/issues/2 
+      # See: https://github.com/openmicroscopy/ansible-role-omego/issues/2
       omero_omego_version: 0.6.4
 
     - role: openmicroscopy.omero-web
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -163,7 +163,6 @@
         - "omero-webtagging-autotag"
         - "omero-webtagging-tagsearch"
         - "omero-iviewer"
-        - "omero-marshal==0.5.1"
         editable: False
         state: "{{ webapps_state }}"
         # variable comes from role openmicroscopy.omero-web
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -135,7 +135,7 @@
       omero_web_release: 5.4.1
 
     # This role only works on OMERO 5.3+
-    - role: omero-user
+    - role: openmicroscopy.omero-user
       no_log: true
       omero_user_bin_omero: /opt/omero/server/OMERO.server/bin/omero
       omero_user_system: omero-server
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -126,6 +126,9 @@
       omero_server_dbpassword: "{{ vault.omero_server_dbpassword }}"
       omero_server_rootpassword: "{{ vault.omero_server_rootpassword }}"
       omero_server_systemd_limit_nofile: 16384
+      # Once omego version bug is fixed, we can remove this hard-coding.
+      # It should be defined in the omero-server role.
+      # See: https://github.com/openmicroscopy/ansible-role-omego/issues/2 
       omero_omego_version: 0.6.4
 
     - role: openmicroscopy.omero-web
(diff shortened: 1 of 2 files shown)


requirements.yml
```diff
@@ -16,7 +16,7 @@
 
 - name: openmicroscopy.omero-server
   src:  https://github.com/openmicroscopy/ansible-role-omero-server.git
-  version: 2.0.0-m1
+  version: 2.0.1
```
sbesson: Now `2.0.2` is out, should this be bumped as I understand the playbook was deployed with the role changes integrated in https://github.com/openmicroscopy/ansible-role-omero-server/pull/22?

kennethgillen:
The playbook at https://github.com/openmicroscopy/prod-playbooks/pull/17/commits/0b80f1912b36bb6bb121155c1fff67aecf69f2c6 upgraded ome-demoserver to 5.4.1 with a modified version of 2.0.1 of ansible-role-omero-server matching 2.0.2 https://galaxy.ansible.com/openmicroscopy/omero-server/

ome-demoserver.yml
```diff
@@ -126,10 +126,13 @@
       omero_server_dbpassword: "{{ vault.omero_server_dbpassword }}"
       omero_server_rootpassword: "{{ vault.omero_server_rootpassword }}"
       omero_server_systemd_limit_nofile: 16384
+      # Once omego version bug is fixed, we can remove this hard-coding.
```
joshmoore: Perhaps add a `TODO`

kennethgillen: Do you mean something more than just adding TODO do the comment? (a la --Fixes in github?) e.g. something for SCC?

joshmoore: Perfect. Thanks.

kennethgillen:
Is it worth figuring out the syntax for https://github.com/openmicroscopy/ansible-role-postgresql/tree/3.0.1 in this PR?

It would mean working out how to use the new variables (`postgresql_users_databases` was split).

I don't see a reason to bump it, other than to keep up to date with the roles / or test the roles on our production infrastructure.

requirements.yml
```diff
@@ -16,7 +16,7 @@
 
 - name: openmicroscopy.omero-server
   src:  https://github.com/openmicroscopy/ansible-role-omero-server.git
```
kennethgillen: Should these all come from Galaxy, rather than GH now?

manics: Yes

joshmoore: Are you pushing this, @kennethgillen ?

kennethgillen: Yes, will pull the GH urls. (Does it apply to all requirements?)

kennethgillen: Ah, @manics already replied to my first question which was "all". Pushing.

kennethgillen:
```
 [WARNING]: - omero-user was NOT installed successfully: Role has no field
named u'owner'
```

kennethgillen:
@manics  - one "switch to galaxy" too far?

joshmoore:
Missing `openmicroscopy.` prefix.

kennethgillen:
d'oh. Thanks.

manics:
Can you match the syntax in the rest of the requirements file: `src:` instead of `name:`?

kennethgillen:
Do you/we prefer `src` or `name`?

```
LS27172:prod-playbooks kenny$ cat requirements.yml  | grep -E "src|name" | awk '{print $2}' | sort | uniq -c
   7 name:
   3 src:
```

sbesson:
From http://docs.ansible.com/ansible/latest/galaxy.html#installing-multiple-roles-from-a-file, `src` is required and `name` is optional. Since we should have fixed the Galaxy names, I am in favor of unifying towards `src: openmicroscopy.<role>` whenever possible.

Re style guide, :+1: I'm slowly going through our contributing documentation and Ansible is definitely one of my upcoming languages if no-one wants to take a first stab.

sbesson:
Thanks. With the last changes, I think we brought these playbooks to the latest known state of the Ansible roles. If no objection from @joshmoore, I would propose we get this and the associated inventory PR in and mark the completton of the prod servers upgrade to 5.4.1 so that we can focus on the next tasks.

joshmoore:
:+1: for getting this in as is. I lost track a bit of the `omero-marshal` discussion, so not sure if anything needs to be captured from that discussion.

sbesson:
My understanding is that at the level of our prod OMERO playbooks, we are currently leaving `omero-marshal`  fully managed alongside the other OMERO.web dependencies via the `requirements-py27.txt` file shipped with the bundle. Concretely this means upgrades will be triggered by version bump in the upstream code - see https://github.com/openmicroscopy/openmicroscopy/pull/5595.

We can re-evaluate this strategy when we have a need for a redeployment of the Web client with a  a new version of `omero-marshal`.

kennethgillen:
@sbesson - there are some newer roles such as the PostgreSQL role, but that'll take either input from @manics or some testing time by me to work how how to set the variables to end up with the configuration we want.

sbesson:
Understood. Including these as they are then.

PR #20, opened 2017-12-08
lvm tag for playbook (storage) run isolation

kennethgillen:
Adding lvm tag to playbook, as done in ome-demoserver.yml at https://github.com/openmicroscopy/prod-playbooks/blob/master/ome-demoserver.yml#L27

This allows us to run only the storage part of the playbook when storage changes are required (potentially an on-line operation) without affecting the rest of the system.

### DIFF
--- a/nightshade-webclients.yml
+++ b/nightshade-webclients.yml
@@ -20,6 +20,7 @@
 
     # Root LV Size
     - role: openmicroscopy.lvm-partition
+      tags: lvm
       lvm_lvname: "{{ provision_root_lvname }}"
       lvm_vgname: "{{ provision_root_vgname }}"
       lvm_lvmount: /


sbesson:
As the `lvm` tag is already defined elsewhere, re-using it in this playbook makes sense.

As a side discussion, an alternate strategy would be to split the group of tasks/roles that need to be run in isolation into sub-playbooks (see https://github.com/openmicroscopy/prod-playbooks/pull/5 or https://github.com/IDR/deployment).

joshmoore:
WFM. We can review the overall strategy later, deprecating tags, etc as needed.

manics:
Could we at least start a doc in this repo listing tags, or add them to the README? e.g. like https://github.com/kubernetes-incubator/kubespray/blob/v2.3.0/docs/ansible.md#ansible-tags

joshmoore:
Nice list!

kennethgillen:
Are you keeping track of Kubernetes tags like that, too?

manics:
Since you've now looked at that list... one thing to consider is how to test tags in isolation, and how to ensure you don't end up with an inconsistent state (e.g. from experience kubespray has some undocumented dependencies between tags)

manics:
> Are you keeping track of Kubernetes tags like that, too?

What Kubernetes tags?

joshmoore:
Re: testing -- agreed. Whether we use a tag or a separate playbook strategy, we'll need to agree on the semantics and what's safe when. e.g. when it comes to www/jekyll.yml, I've *only* been using the www-jekyll.yml sub-playbook, so the rest is getting less testing. (The reverse of what @manics is pointing out)

kennethgillen:
How about an additional tag for runtime safety. e.g.  `--on-line`. We can generate a new column on our version of that list from the presence/absence of this tag.

joshmoore:
See https://github.com/openmicroscopy/prod-playbooks/issues/24

joshmoore:
SA call today had no vetoes. #24 captures any remaining issues.

PR #25, opened 2017-12-18
Bump openmicroscopy.omego version to 0.1.1

sbesson:
Also drop omego versions overriding in the server playbooks


kennethgillen:
Thanks, @sbesson.

Issue #21, opened 2017-12-11
use of pip module means "--check" always shows `changed`


kennethgillen:
use of pip module means "--check"mode always shows `changed` for pip tasks.

Internal OME links: 
 https://openmicroscopy.slack.com/archives/C0K5ME1EW/p1512749355000430 
 https://github.com/openmicroscopy/management_tools/pull/612

manics:
Has anyone tested with the latest ansible 2.4.2 (or even 2.4.3.0-0.1.beta1)?

manics:
Also is this an issue only with check mode (so when running for real you **don't** get incorrect statuses or handlers being trigger)?

kennethgillen:
@manics re: your last: I haven't re-run since the demo upgrade (which had to restart web anyway) so can't say. We'll have to make a note next time we run these.

sbesson:
My recent run of the demo-server using Ansible 2.3.1.0 without the check mode showed no changed state for the `openmicroscopy.omero-web : omero web | setup virtualenv` task.

manics:
Is this still relevant?

joshmoore:
Yes. I'm certainly still periodically seeing `changed` for pip lines when using check mode

PR #26, opened 2017-12-19
training-server: migrate from examples

joshmoore:
See: https://github.com/ome/ansible-examples-omero/pull/8

NB: this description will be updated with testing instructions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ env:
   - DIRECTORY="." PLAYBOOK=nightshade-webclients.yml
   - DIRECTORY="." PLAYBOOK=ome-demoserver.yml
   - DIRECTORY="." PLAYBOOK=ome-dundeeomero.yml
+  - DIRECTORY="." PLAYBOOK=training-server.yml
   - DIRECTORY=bootstrap PLAYBOOK=
   - DIRECTORY=web-proxy PLAYBOOK=playbook.yml
   - DIRECTORY=www PLAYBOOK=""
--- a/training-server.yml
+++ b/training-server.yml
@@ -1,93 +1,156 @@
-# Production tuning of server limits.conf, Postgres and NGINX required, but this will install a minimal OMERO server,
-# OMERO.web with NGINX and systemd configuration controlling both.
+# Install OMERO.server and OMERO.web with a public user on localhost
 
-# Install the latest release, including PostgreSQL on the same server
+- hosts: ome-training-outreach.openmicroscopy.org
+  pre_tasks:
 
-# users-to-be-added to come from host_vars
+    - name: Install Make Movie script Prerequisite | MEncoder - Repo
+      become: yes
+      yum:
+        name:  http://li.nux.ro/download/nux/dextop/el7/x86_64/nux-dextop-release-0-5.el7.nux.noarch.rpm
+        state: present
 
-- hosts: training-server
+    - name: Install Make Movie script Prerequisite | MEncoder - Package
+      become: yes
+      yum:
+        name: mencoder
+        state: present
+
+    - name: OMERO.figure server-side prerequisites, script prerequisites + web server for decoupled OMERO.web
+      become: yes
+      yum:
+        name: "{{ item }}"
+        state: present
+      with_items:
+        # For OMERO.figure
+        - python-reportlab
+        - python-markdown
+        # For the 'make movie' script
+        - mencoder
+        - scipy
+
   roles:
-  - role: omero-server
-    # Taking this from the openstack playbook os-omero.yml, to install the systemd config.
-    omero_systemd_setup: True
+
+    - role: openmicroscopy.postgresql
+      postgresql_databases:
+      - name: omero
+      postgresql_users:
+      - user: omero
+        password: omero
+        databases: [omero]
+      postgresql_version: "9.6"
+
+    - role: openmicroscopy.omero-server
+      omero_server_release: 5.4.1
+
+    - role: openmicroscopy.omero-web
+      omero_web_release: 5.4.1
+      # omero_web_config_set:
+
+  post_tasks:
+    - name: Omero.web plugins | plugin install via pip & pypi
+      become: yes
+      pip:
+        name:
+        - "omero-figure==3.1.2"
+        - "omero-fpbioimage==0.2.0"
+        - "omero-webtagging-autotag==3.0.2"
+        - "omero-webtagging-tagsearch==3.0.3"
+        - "omero-iviewer==0.4.1"
+        editable: False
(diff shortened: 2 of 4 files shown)


joshmoore:
Check-mode run at this point shows:

```
# (v2)joshuas-mbp:prod jamooreansible-playbook playbooks/site.yml -l outreach.openmicroscopy.org -i ~/opt/management_tools/ansible/inventory/prod-hosts -C --diff
...
TASK [Config for OMERO.web plugins] ********************************************************************************************************************************************************************************************************************************************
ok: [outreach.openmicroscopy.org]

TASK [Set default viewer in OMERO.web] *****************************************************************************************************************************************************************************************************************************************
ok: [outreach.openmicroscopy.org]

RUNNING HANDLER [openmicroscopy.omero-common : restart omero-web] **************************************************************************************************************************************************************************************************************
changed: [outreach.openmicroscopy.org]

PLAY RECAP *********************************************************************************************************************************************************************************************************************************************************************
outreach.openmicroscopy.org : ok=84   changed=6    unreachable=0    failed=0

```

i.e. a small number of (spurious) pip changes and a web restart. But looks generally good.

sbesson:
Some of the spurious pip changes might be related to the usage of `check-mode` - cf https://github.com/openmicroscopy/prod-playbooks/issues/21. Otherwise looks good to me and ready to merge assuming @kennethgillen and @pwalczysko are happy

kennethgillen:
👍 - thanks @pwalczysko, great addition.

training-server.yml
```diff
@@ -1,93 +1,156 @@
...
 
-# Install the latest release, including PostgreSQL on the same server
+- hosts: outreach.openmicroscopy.org
+  pre_tasks:
 
-# users-to-be-added to come from host_vars
+    - name: Install Make Movie script Prerequisite | MEncoder - Repo
+      become: yes
+      yum:
+        name:  http://li.nux.ro/download/nux/dextop/el7/x86_64/nux-dextop-release-0-5.el7.nux.noarch.rpm
```
jburel: i remember having issue with that url
Is it more stable now?

training-server.yml
```diff
@@ -1,93 +1,156 @@
...
+      yum:
+        name: "{{ item }}"
+        state: present
+      with_items:
+        # For OMERO.figure
+        - python-reportlab
+        - python-markdown
+        # For the 'make movie' script
+        - mencoder
+        - scipy
```
jburel: the comment is a bit ambiguous
scipy is not for make movie but for another script (not official one yet)

training-server.yml
pwalczysko: the playbook was run more than 20 times in last 2 months. Never any issue occured with this url.

manics: Also it's not https, do we still want to publicise this?

training-server.yml
```diff
@@ -1,93 +1,156 @@
-# Production tuning of server limits.conf, Postgres and NGINX required, but this will install a minimal OMERO server,
-# OMERO.web with NGINX and systemd configuration controlling both.
+# Install OMERO.server and OMERO.web with a public user on localhost
 
-# Install the latest release, including PostgreSQL on the same server
+- hosts: outreach.openmicroscopy.org
+  pre_tasks:
```
manics: Minor point, but do these need to be `pre_tasks` (implying they're prerequisites for roles) instead of `post_tasks`?

training-server.yml
```diff
@@ -1,93 +1,156 @@
...
+        state: directory
+        mode: 0755
+        recurse: yes
+        owner: "omero-server"
+        group: "omero-server"
+
+    - name: Download the Scipy_Gaussian_Filter.py script
+      become: yes
+      get_url:
+        url: https://raw.githubusercontent.com/jburel/omero-example-scripts/cambridge-training/cambridge-training-2017-12/Scipy_Gaussian_Filter.py
```
manics: I guess this will (eventually) end up in an org repo?

.travis.yml
```diff
@@ -29,6 +29,7 @@ env:
   - DIRECTORY="." PLAYBOOK=nightshade-webclients.yml
   - DIRECTORY="." PLAYBOOK=ome-demoserver.yml
   - DIRECTORY="." PLAYBOOK=ome-dundeeomero.yml
+  - DIRECTORY="." PLAYBOOK=training-server.yml
```
manics: It looks like `training-server.yml` doesn't have any hardware/low-level config, so is a candidate for a full molecule test instead of just `molecule-syntaxcheck.yml`

training-server.yml
kennethgillen: @jburel - I've never noticed a failure at run-time, either. 

@manics - [we already have](https://github.com/openmicroscopy/prod-playbooks/search?utf8=%E2%9C%93&q=http%3A%2F%2Fli.nux.ro%2Fdownload&type=).  If you've got a better suggestion, we're all 👂👂

training-server.yml
kennethgillen: @jburel - happy to have this updated on all the playbooks where it's used. The comment was really for the line directly below it, but sure, that's not clear. Can we add comments on the same line, in YAML, @manics? i.e. append the comment after `mencoder`?

kennethgillen:
We'll probably want another PR for all of these comments/reviews ^

training-server.yml
kennethgillen: Aha, yes, we can! https://learnxinyminutes.com/docs/yaml/

jburel: so far we have only used that script for training
very minor comment

jburel:
@manics could you open a follow-up PR please?

training-server.yml
jburel: it will 
unfortunately we did not have time to set repo etc.
too many moving parts

training-server.yml
jburel: I have not found any repo offering https
getting rid of mencoder is probably the solution
Note that I have not been able to test the movie script via travis

training-server.yml
kennethgillen: @jburel  - https://github.com/openmicroscopy/prod-playbooks/pull/27

training-server.yml
pwalczysko: Getting rid of mencoder --> the movie script stops to work --> we are going to an outreach with one installed script which is not working

Do I get this right ? Does not seem to me as a good solution.

jburel: @pwalczysko sorry for the confusion. I meant using a more recent library for encoding
if we want to keep such feature but this will imply that we have to rewrite portion of the script
For nowI think the url used is the best we have found

PR #27, opened 2017-12-20
prod-playbooks: Adjusted comments

kennethgillen:
Adjusted the comments around the `yum` install tasks to be more specific.

### DIFF
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -53,8 +53,7 @@
         # For OMERO.figure
         - python-reportlab
         - python-markdown
-        # For the 'make movie' script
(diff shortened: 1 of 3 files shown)


kennethgillen:
Note: 

```
ansible-lint training-server.yml
[ANSIBLE0015] Found a bare variable 'start=1 end=40 format=user-%d' used in a 'with_sequence' loop. You should use the full variable syntax ('{{start=1 end=40 format=user-%d}}')
training-server.yml:114
Task/Handler: Add DropBox folders for user-1 to user-40
```

joshmoore:
:+1:

PR #28, opened 2018-01-03
Add nginx-proxy 1.6.0

manics:
Add nginx-proxy 1.6.0


sbesson:
The role version bump makes sense to benefit from the improvements of https://github.com/openmicroscopy/ansible-role-nginx-proxy/pull/15. My primary concern here is the declaration of two separate versions of `nginx-proxy` role in `www/requirements.yml` and in the top-level `requirements.yml`. Ideally, I would vote for unifying these two requirements files or minimally have an agreement on the level of divergence we would accept.

manics:
I agree, however I that's quite a big change for www and should be handled separately.

manics:
The downside of requiring all systems to use the same versions of all roles is you've effectively coupled their deployments together. One consequence of this is they all have to follow the same deployment pattern, e.g. rapid updates and deployments vs infrequent scheduled updates, so we'd need to agree on this.

manics:
The last commit merges www/requirements.yml into requirements.yml. The only discrepancy is the new `openmicroscopy.nginx-proxy` version which I've named `openmicroscopy.nginx-proxy-1.6.0` to allow both version to coexist.

kennethgillen:
I had similar concerns as https://github.com/openmicroscopy/prod-playbooks/pull/28#issuecomment-355238073 when the prod-playbooks were refactored to remove the requirements.yml I had alongside each playbook file (individual server state per playbook).

sbesson:
Keeping in mind the concerns expressed above about maintaining a single requirements file for all server state playbooks, one advantage of 7a724b is to explicitly declare all the roles used across the playbooks incl. those used at different versions and minimize potential conflicts/stale installations under `vendor/`.
I propose we use tomorrow's weekly meeting to sign off on it with @openmicroscopy/devops (or agree on an alternate strategy), then I will test the deployment.

manics:
Deployed on web-proxy

joshmoore:
Merging as the current state of production. Leaving someone else to best record the cleanup step.

PR #33, opened 2018-01-25
Downgrade OMERO.web for NS to 5.4.1

kennethgillen:
Downgraded Nightshade OMERO.web servers to 5.4.1.

Note: since the Server is at 5.4.2, you must set the following config on your OMERO.web instances to allow them to communicate with an older server.

`bin/omero config set omero.web.check_version false`

Part of https://github.com/openmicroscopy/management_tools/issues/642

Deployed on ns-web and ns-web-pub on 20180125


joshmoore:
Merging for a direct bump to 5.4.3

PR #36, opened 2018-01-30
Adding necessary managed repo config

kennethgillen:
Since the role no longer pulls in the `omero_server_datadir` defined in `host_vars`.


manics:
Looks OK.

As mentioned on https://github.com/openmicroscopy/ansible-role-omero-server/issues/28 it would be useful to know why `omero_server_datadir_manage` is needed

joshmoore:
Restarted job re: `Failed to connect to galaxy.ansible.com at port 443: [Errno -2] Name or service
`

joshmoore:
Now green. Do we want this in an OMERO-5.4.3 tag or is there discussion still needed?

manics:
It's deployed so I guess it should be regardless of discussions.

joshmoore:
Fair enough. Merging and tagging then.

PR #37, opened 2018-01-31
Requirements unification and cleanup

sbesson:
Follow-up of https://github.com/openmicroscopy/prod-playbooks/pull/28

## Changes

- unifies the version of `nginx-proxy` to  1.7.0
- bumps the version of `jeykll-build` to 2.0.0 to fix the Travis build
- removes the post-task modifying the nginx worker processes as already handled via the `nginx_proxy_worker_processes` parameter
- removes the post-task modifying nginx worker connection to use the new `nginx_proxy_worker_connections` parameter introduced in 1.7.0

## Test

The `jekyll-build` bump should fix Travis which should pass for all playbooks.
In terms of deployment, the following playbooks should be tested with and without check-mode:

- ```$ ansible-playbook -i /path/to/ansible/inventory/prod-hosts --ask-become -C --diff playbooks/web-proxy/playbook.yml```
- ```$ ansible-playbook -i ../ansible/inventory/staging-hosts --ask-become -C --diff playbooks/www/www-deploy.yml```:

In both cases, nginx changes should only result in whitespace changes while the jekyll-build changes should modify the base ruby installation to use `rvm`. This PR should have no functional modification.

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,7 +1,7 @@
 ---
 
 - src: rvm_io.ruby
-  version: 2.0.1
+  version: v2.0.1
 
 - name: openmicroscopy.jekyll-build
   src: https://github.com/sbesson/ansible-role-jekyll-build/archive/rvm_dependency.tar.gz
--- a/requirements.yml
+++ b/requirements.yml
@@ -9,8 +9,7 @@
 - src: openmicroscopy.network
   version: 1.0.0
 
-- name: openmicroscopy.nginx-proxy
-  src: https://github.com/sbesson/ansible-role-nginx-proxy/archive/worker_connections.tar.gz
+- src: openmicroscopy.nginx-proxy
   version: 1.7.0
 
 - src: openmicroscopy.nginx-ssl-selfsigned
--- a/requirements.yml
+++ b/requirements.yml
@@ -9,8 +9,9 @@
 - src: openmicroscopy.network
   version: 1.0.0
 
-- src: openmicroscopy.nginx-proxy
-  version: 1.6.0
+- name: openmicroscopy.nginx-proxy
+  src: https://github.com/sbesson/ansible-role-nginx-proxy/archive/worker_connections.tar.gz
+  version: 1.7.0
 
 - src: openmicroscopy.nginx-ssl-selfsigned
   version: 1.0.0
--- a/requirements.yml
+++ b/requirements.yml
@@ -10,10 +10,6 @@
   version: 1.0.0
 
 - src: openmicroscopy.nginx-proxy
-  version: 1.5.2
-
-- name: openmicroscopy.nginx-proxy-1.6.0
-  src: openmicroscopy.nginx-proxy
   version: 1.6.0
 
 - src: openmicroscopy.nginx-ssl-selfsigned
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,7 +1,8 @@
 ---
 
-- src: openmicroscopy.jekyll-build
-  version: 1.2.1
+- name: openmicroscopy.jekyll-build
+  src: https://github.com/sbesson/ansible-role-jekyll-build/archive/rvm_dependency.tar.gz
+  version: 2.0.0
 
 - src: openmicroscopy.lvm-partition
   version: 1.1.0
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,8 +1,11 @@
 ---
 
+- src: rvm_io.ruby
+  version: 2.0.1
+
 - name: openmicroscopy.jekyll-build
   src: https://github.com/sbesson/ansible-role-jekyll-build/archive/rvm_dependency.tar.gz
-  version: 2.0.0
+  version: 1.3.0
 
 - src: openmicroscopy.lvm-partition
   version: 1.1.0
--- a/requirements.yml
+++ b/requirements.yml
@@ -3,8 +3,7 @@
 - src: rvm_io.ruby
   version: v2.0.1
 
-- name: openmicroscopy.jekyll-build
-  src: https://github.com/sbesson/ansible-role-jekyll-build/archive/rvm_dependency.tar.gz
+- src: openmicroscopy.jekyll-build
   version: 1.3.0
 
 - src: openmicroscopy.lvm-partition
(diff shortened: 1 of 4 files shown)


kennethgillen:
@sbesson - it looks like `nginx_proxy_worker_processes` is already in https://github.com/openmicroscopy/prod-playbooks/blob/master/www/www-deploy.yml#L22 ?

So if that's already covered, then great. Probably https://github.com/openmicroscopy/prod-playbooks/blob/master/www/www-deploy.yml#L34-L40 should have been removed when ^ was added.

kennethgillen:
@sbesson  - I would say it makes total sense to PR the role to accept connections in the same way.

sbesson:
Pushed another commit consuming https://github.com/openmicroscopy/ansible-role-nginx-proxy/pull/16 and ran the two playbooks mentioned in the description in check-mode. Will list these for review.

sbesson:
Last 2 commits also consume the `openmicroscopy.jekyll-build` changes proposed in https://github.com/openmicroscopy/ansible-role-jekyll-build/pull/10 allowing to fix the Travis build. 

Unfortunately, running it in check-mode against `ome-www-dev` failed with:

```
TASK [rvm_io.ruby : Symlink ruby related binaries on the system path] **********
failed: [ome-www-dev.openmicroscopy.org] (item=erb) => {"failed": true, "item": "erb", "msg": "src file does not exist, use \"force=yes\" if you really want to create the link: /usr/local/rvm/wrappers/default/erb", "path": "/usr/local/bin/erb", "src": "/usr/local/rvm/wrappers/default/erb", "state": "absent"}
```

Given validation by the Travis, we might still want to run this against `ome-www-dev` outside check-mode (after snapshotting).

joshmoore:
Nicely green. As I understand it, the primary discussion is around whether we're ok with `pre_tasks` (knowing that we didn't have the best of luck with `post_tasks`) and whether or not the `rvm` strategy is generally acceptable, correct?

Looking around, the most definitive statement I found was on https://github.com/ansible/ansible/issues/3869#issuecomment-22761648 (2013):

> Pre tasks and post tasks are largely designed for working with load balancers.

which is matched by most of the doc references I could find being on http://docs.ansible.com/ansible/latest/guide_rolling_upgrade.html 

Not saying it shouldn't be used right now, but it may be something we want to re-consider. I assume the "wrapper role" suggestion came from https://github.com/ansible/ansible/issues/13228. That certainly looks _less_ appealing than what you've done in jekyll-build#10

sbesson:
> As I understand it, the primary discussion is around whether we're ok with pre_tasks (knowing that we didn't have the best of luck with post_tasks) and whether or not the rvm strategy is generally acceptable, correct?

Yes

> Not saying it shouldn't be used right now, but it may be something we want to re-consider. I assume the "wrapper role" suggestion came from ansible/ansible#13228. 

Yes, the suggestion came from this issue as well as other threads like [this one](https://stackoverflow.com/questions/29258759/how-to-specify-ansible-pretasks-for-a-role) discussing pretasks for roles.

Another alternative to the `pre_tasks` usage precluding the wrapper role strategy would be to use  the [include_role](http://docs.ansible.com/ansible/latest/include_role_module.html) module ad install `which` and `rvm` within `tasks/main.yml`. Here again, caveats apply since `This module is flagged as preview which means that it is not guaranteed to have a backwards compatible interface.` but I can try to have a secondary branch in `jekyll-build` implementing this strategy so that we can make our decision knowing the status of all the options.

joshmoore:
Another alternative would be to fork the rvm role and to add a task to install `which`, correct?

sbesson:
https://github.com/openmicroscopy/ansible-role-jekyll-build/compare/master...sbesson:rvm_dependency_2 implements the `include_role` and seems to be suitable as well (passing the `molecule` tests).

Tried to summarize the available options in preparation of the discussion at the upcoming meeting

| Strategy | jekyll-build role changes                                                                                          | www playbook changes                       | Pros                                                                                    | Cons                                                       |
|----------|--------------------------------------------------------------------------------------------------------------------|--------------------------------------------|-----------------------------------------------------------------------------------------|------------------------------------------------------------|
| A        | jekyll-build 1.2.0: add role dependency on `rvm_io.ruby` and install `which` package as a `pre_task` on CentOS     | bump to jekyll-build:1.2.0                 | Fixes Travis using upstream Galaxy role                                                 | Relies on `pre_task` which was not designed for this       |
| B        | jekyll-build 1.2.0: add `rvm_io.ruby` as a role task using `include_role`                                          | bump to jekyll-build:1.2.0                 | Fixes Travis using upstream Galaxy role                                                 | Relies on include_role which is a preview module           |
| C        | jekyll-build 1.2.0: fork `rvm_io.ruby` as `ome.ruby`, install `which` as a task, add role dependency on `ome.ruby` | bump to jekyll-build:1.2.0                 | Uses basic role semantics. Can be backported to the upstream role.                      | Burden of maintaining a fork.                          |
| D        | none                                                                                                               | Rewrite to unpack bundle of static website | Probably the best long-term strategy. Removes ruby/jekyll dependency on the www server. | Requires investigation. Disable `www` playbooks in Travis. |

sbesson:
Discussed with @openmicroscopy/devops. Consensus was to go for solution B i.e. `include_role` and deploy on `ome-www-dev` for acceptance. Separately, will try to open a PR adding support in the upstream role for the `which` installation and card the investigation of the website bundle creation and deployments.

sbesson:
With the last changes from https://github.com/openmicroscopy/ansible-role-jekyll-build/pull/10, this has been successfully run against `ome-www-dev.openmicroscopy.org`. Automatic and manual tests seem to indicate everything remains functional. Any reason not to move forward with the release of the `jekyll-build` role and the integration of this requirements unification PR?

kennethgillen:
Did someone test `web-proxy` as outlined ^ at the start?

sbesson:
With the last commit, proposing to merge this, restart Travis for all open PRs against prod-playbooks and run the `www` and `ome-www` playbooks /cc @openmicroscopy/devops

joshmoore:
:+1: I'll test with gh-42 now.

sbesson:
See https://github.com/rvm/rvm1-ansible/pull/165 for the proposed upstream change.

PR #38, opened 2018-02-07
Use shared version vars for omero-figure

manics:
Refactors nightshade-webclients.yml ome-demoserver.yml ome-dundeeomero.yml to use a common set of optional variables for OMERO.figure.


jburel:
Tested the changes in checkmode this morning and everything works as expected

PR #40, opened 2018-02-12
Add OMERO features redirects for new navigation

hflynn:
Adds redirects needed by https://github.com/openmicroscopy/www.openmicroscopy.org/pull/195

### DIFF
--- a/www/www-deploy.yml
+++ b/www/www-deploy.yml
@@ -143,23 +143,23 @@
       dest: https://www-legacy.openmicroscopy.org/site/products/$link
 
     # OMERO features
-    - match: "~/omero/new/"
+    - match: "~/omero/new/?$"
       dest: /omero/features/
-    - match: "~/omero/import/"
+    - match: "~/omero/import/?$"
       dest: /omero/features/import/
-    - match: "~/omero/organize/"
+    - match: "~/omero/organize/?$"
       dest: /omero/features/organize/
-    - match: "~/omero/view/"
+    - match: "~/omero/view/?$"
       dest: /omero/features/view/
-    - match: "~/omero/analyze/"
+    - match: "~/omero/analyze/?$"
       dest: /omero/features/analyze/
-    - match: "~/omero/publish/
+    - match: "~/omero/publish/?$"
       dest: /omero/features/publish/
-    - match: "~/omero/export/"
+    - match: "~/omero/export/?$"
       dest: /omero/features/export/
-    - match: "~/omero/learning/"
+    - match: "~/omero/learning/?$"
       dest: /omero/features/learning/
-    - match: "~/omero/utility/"
+    - match: "~/omero/utility/?$"
       dest: /omero/features/utility/
 
     # community
--- a/www/www-deploy.yml
+++ b/www/www-deploy.yml
@@ -126,7 +126,7 @@
     - match: "~/site/products/omero/downloads/?$"
       dest: /omero/downloads/
     - match: "~/site/products/omero/feature-list/?$"
-      dest: /omero/new
+      dest: /omero/features/
     - match: "~/site/products/omero/big-images-support/?$"
       dest: /omero/view/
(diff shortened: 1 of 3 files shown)


www/www-deploy.yml
```diff
@@ -142,6 +142,26 @@
     - match: "~/site/products/(?<link>.*)$"
       dest: https://www-legacy.openmicroscopy.org/site/products/$link
 
+    # OMERO features
+    - match: "~/omero/new/"
```
sbesson: The `~` i.e. the usage  of regexp,  might be unnecessary here since we are testing exact URLs - unless we want to handle URLs with and without trailing slash i.e. "~/omero/new/?$"

sbesson:
Deployed on `ome-www-dev`. Prior to the deployment, the tests indicated 40 failures. Post deployment, this is reduced to 36. The new `site/` redirect but the `/omero` one are not working. Looking at the diff, if we want to implement this in this playbook, `nginx_proxy_redirect_map_locations`  would need to be expanded to also include these type of URLs.

With a bit of perspective on this, one consideration is that this entire body of work is about redirecting URLs internally. This departs from what these redirect rules are trying to achieve so far as they redirecting legacy URLs. My feeling is that the proper way to implement should be back in the upstream repository using `jekyll-redirect-from` as we previously did for `ome-help`. The tests added to this repository could remain unchanged as they are effectively capturing the state of redirected URLs. Thoughts?

hflynn:
Adding the redirects to https://github.com/openmicroscopy/www.openmicroscopy.org/pull/195 instead and using the jekyll redirections works for me. 

I could remove the new URLs section here and leave the tests and the update to the existing redirect.

hflynn:
@sbesson this should be sorted now - added the redirects to the existing website PR but left them in the tests here

sbesson:
Deployed the last version on `ome-www-dev.openmicroscopy.org` but the redirect tests are failing. My suspicion is that we need to add the additional `gem` to upstream `Gemfile` as the role does not install it from `_config.yml`.

hflynn:
I'll push another commit to the website PR - I didn't touch the Gemfile in the end because locally it worked and I can't actually get my local jekyll to build without deleting my Gemfile due to some Ruby version conflict.

hflynn:
@sbesson updated the website PR now

sbesson:
Deployed the latest version of this playbook using the upstream branch with `jekyll-redirect-from` on `ome-www-dev.openmicroscopy.org`. Functionally, everything works as expected.

Except for `/site/products/omero/feature-list`, the new redirect tests added in this PR are all failing due to the fact that these URLs use a different redirection method based on the `meta` HTML tags rather than nginx. Quick investigation suggests that testing this type of redirects will require more code than the basic `requests` API.

Proposal would be to merge the upstream PR and strip the failing redirects tests from this PR.

hflynn:
@sbesson hopefully sorted now

sbesson:
Upstream PR merged, can you revert the changes in 30d787ee8b629142bfa69ecda2a2982e5d322596 and I will test the final deployment?

hflynn:
@sbesson pushed

sbesson:
```
(ansible) sbesson@ls30630:prod (master/merge/trigger) $ HOST=http://ome-www-dev.openmicroscopy.org pytest playbooks/www/tests/check_redirects.py 
======================================================================== test session starts =========================================================================
platform darwin -- Python 2.7.14, pytest-3.3.2, py-1.5.2, pluggy-0.6.0
rootdir: /private/tmp/management_tools/prod, inifile:
plugins: testinfra-1.5.5, pyfakefs-3.3
collected 234 items                                                                                                                                                  

playbooks/www/tests/check_redirects.py ....................................................................................................................... [ 50%]
...................................................................................................................                                            [100%]

==================================================================== 234 passed in 28.67 seconds =====================================================================
```

and individual pages redirect as expected. Merging and deploying on live website.

PR #32, opened 2018-01-23
Get outreach in sync

pwalczysko:
@joshmoore encouraged me to 

1. move the new changes in ``ansible-examples-omero`` repo to this repo
2. work from now on from this repo rather than ``ansible-examples-omero``

The PR adds some os users needed for outreaches with facility managers who are supposed to ssh into the machine, to perform imports in-place as well as have an easy access to CLI in this way.

Also, it adds the docker install and the ldap installation in the docker.

PR #41, opened 2018-02-12
Get training in sync

pwalczysko:
This is a follow-up of https://github.com/openmicroscopy/prod-playbooks/pull/32.
Tried to run the trainin...yml playbook as described in https://github.com/openmicroscopy/prod-playbooks/pull/32#issuecomment-360505390.

This works, except that the password for ldap is either wrongly passed into [the config](https://github.com/openmicroscopy/prod-playbooks/pull/41/commits/e50167269e6aba6c15067cc718faf3e3f3db6dec#diff-dece4a429f875080d5d590525788c328R231) by me, or it does not work with salting, or both. When I hard-code the password for ldap 

``omero.ldap.password: "$HARDCODED_PWD" ``

then the whole playbook works as expected, even when run from the management repo.

I would not try fixing the password now without consulting @joshmoore , because he did have an opinion about how to do the passwords properly within this repo - @joshmoore would you please remind me how to start there (cannot find the card where we discussed it) ?
 
cc @kennethgillen @manics @sbesson

### DIFF
--- a/training-server.yml
+++ b/training-server.yml
@@ -213,6 +213,13 @@
 
 
   vars:
+    #omero_server_datadir_chown: True
+    omero_server_system_managedrepo_group: managed_repo_group
+    omero_server_datadir_managedrepo_mode: u=rwX,g=srwX,o=rX,+t
+    omero_server_datadir_chown: True
+    os_system_users_password: "{{ os_systems_users_password_override | default('$6$leKi5B1PgSvdA/ec$xbU3CnoSFnYdeZjEjKK5TH8SGATsW746uopssff4edpgyu.cWXGo9A.oK6wH9kIkxLCCNcORGZnnroZPMqGzN/') }}"
+    apache_docker_release: "{{ apache_docker_release_override | default('0.3.0') }}"
+    ldap_password: "{{ ldap_password_override | default ('secret') }}"
     omero_server_config_set:
       #omero.fs.importUsers: "fm1"
       omero.fs.watchDir: "/home/DropBox"
@@ -228,14 +235,8 @@
       omero.ldap.group_mapping: "name=cn"
       omero.ldap.new_user_group: "MyData"
       omero.ldap.new_user_group_owner: "(owner=@{dn})"
-      omero.ldap.password: "{{ ldap_password | default('secret')}}"
+      omero.ldap.password: "{{ ldap_password }}"
       omero.ldap.sync_on_login: "true"
       omero.ldap.user_filter: "(objectClass=person)"
       omero.ldap.user_mapping: "omeName=uid,firstName=givenName,lastName=sn,email=mail"
       omero.ldap.username: "uid=admin,ou=system"
-    #omero_server_datadir_chown: True
-    omero_server_system_managedrepo_group: managed_repo_group
-    omero_server_datadir_managedrepo_mode: u=rwX,g=srwX,o=rX,+t
-    omero_server_datadir_chown: True
-    os_system_users_password: $6$leKi5B1PgSvdA/ec$xbU3CnoSFnYdeZjEjKK5TH8SGATsW746uopssff4edpgyu.cWXGo9A.oK6wH9kIkxLCCNcORGZnnroZPMqGzN/
-    apache_docker_release: 0.3.0
--- a/training-server.yml
+++ b/training-server.yml
@@ -135,7 +135,7 @@
     - name: Download the ldap scripts
       become: yes
       get_url:
-        url: https://raw.githubusercontent.com/openmicroscopy/apacheds-docker/master/bin/ldapmanager
+        url: https://raw.githubusercontent.com/openmicroscopy/apacheds-docker/{{ apache_docker_release }}/bin/ldapmanager
         dest: /home/ldap/ldapmanager
         mode: 0755
         force: yes
@@ -204,7 +204,7 @@
     - name: Run docker for ldap
       become: yes
       docker_container:
-        image: openmicroscopy/apacheds
+        image: openmicroscopy/apacheds:{{ apache_docker_release }}
         name: ldap
         published_ports:
         - "10389:10389"
@@ -228,7 +228,7 @@
       omero.ldap.group_mapping: "name=cn"
       omero.ldap.new_user_group: "MyData"
       omero.ldap.new_user_group_owner: "(owner=@{dn})"
-      omero.ldap.password: "{{ ldap_password | default(secret)}}"
+      omero.ldap.password: "{{ ldap_password | default('secret')}}"
       omero.ldap.sync_on_login: "true"
       omero.ldap.user_filter: "(objectClass=person)"
       omero.ldap.user_mapping: "omeName=uid,firstName=givenName,lastName=sn,email=mail"
--- a/training-server.yml
+++ b/training-server.yml
@@ -1,6 +1,6 @@
 # Install OMERO.server and OMERO.web with a public user on localhost
 
-- hosts: all
+- hosts: outreach.openmicroscopy.org
   pre_tasks:
 
     - name: Install Make Movie script Prerequisite | MEncoder - Repo
--- a/training-server.yml
+++ b/training-server.yml
@@ -237,3 +237,5 @@
     omero_server_system_managedrepo_group: managed_repo_group
     omero_server_datadir_managedrepo_mode: u=rwX,g=srwX,o=rX,+t
     omero_server_datadir_chown: True
+    os_system_users_password: $6$leKi5B1PgSvdA/ec$xbU3CnoSFnYdeZjEjKK5TH8SGATsW746uopssff4edpgyu.cWXGo9A.oK6wH9kIkxLCCNcORGZnnroZPMqGzN/
+    apache_docker_release: 0.3.0
--- a/training-server.yml
+++ b/training-server.yml
@@ -1,4 +1,4 @@
-# Install OMERO.server and OMERO.web with a public user on localhost
+# Install OMERO.server and OMERO.web on localhost
 
 - hosts: outreach.openmicroscopy.org
   pre_tasks:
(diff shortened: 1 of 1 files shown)


joshmoore:
I would suggest that:

 * an `outreach.openmicroscopy.org.yml` file be added to the `ansible/inventory/host_vars` directory of management_tools;
 * it have a line `ldap_password: YOURPASSWORDHERE` (which I assume can be in plain text);
 * the section under config_set be updated to: `omero.ldap.password: "{{ ldap_password | default(secret)}}"`; and
 * and _possibly_ that the `ldap_password` be set _before_ `omero_server_config_set` (though I'm not sure if this is causing you issues)

pwalczysko:
>omero.ldap.password: "{{ ldap_password | default(secret)}}

This would kind of assume that I can/know how to set another password than the default in the ldap itself. But I do not -> I would end up with the default password being encoded in this roundabout way https://github.com/openmicroscopy/prod-playbooks/pull/41#issuecomment-365182258.

joshmoore:
My goal with the "roundabout way" is to have a playbook that works even when someone does not know our secrets.

pwalczysko:
>and possibly that the ldap_password be set before omero_server_config_set (though I'm not sure if this is causing you issues)

Sorry, I do not understand this instruction. There is no other mention of ``ldap_password`` in the playbook except the places which you mention in the previous instruction points. Thus, there is nothing to ``set``.

pwalczysko:
Is this idea https://github.com/openmicroscopy/prod-playbooks/pull/41#issuecomment-365182258 valid also for the ``"{{ os_system_users_password | default(DEFAULT) }}" ? Atm it is done with the (local ?) salting.

joshmoore:
Yes, this would be my suggestion for how all sensitive variables are transmitted from "elsewhere" to into a public playbook.

pwalczysko:
It looks like I have a working solution for the ldap password as described in https://github.com/openmicroscopy/prod-playbooks/pull/41#issuecomment-365182258. But for the os system users, the analogous does not want to work. This means, my os users now get rejected with 


```
fm1@o...py.org's password: 
Permission denied, please try again.
```

joshmoore:
Is the code available somewhere? If not, is the `os_system` password still crypted (and the `ldap_password` one _not_)? That would be my assumption since the `user:` module assumes crypt'd passwords: http://docs.ansible.com/ansible/latest/user_module.html

pwalczysko:
>since the user: module assumes crypt'd passwords:

You got it. And I, unwisely, am trying to feed it a non-crypted one. This is a strong reason why this should fail.

Attempting to fix this, I thought I would go away from the crypted password even in the user module -> but this seems like a contradiction, as it does not seem possible.
So by now, I have to stick to crypting (which I think is happening only locally) when using the user module ?

joshmoore:
This sounds correct and is not too onerous. If you would prefer to have the system password in plaintext rather than encrypted, I imagine we could come up with a small example that does so in the playbook itself, e.g. https://unix.stackexchange.com/questions/273316/add-user-and-set-password-using-ansible

pwalczysko:
Added the changes for both ldap and os users' passwords as discussed. Should I be opening a PR in management repository for my changes in the new ``outreach...py.org.yml`` file ?

I suspect that the ``bundle`` problem in travis here will repeat as I did not do anything against it. Any ideas please ?

joshmoore:
> Should I be opening a PR in management repository for my changes in the new outreach...py.org.yml file ?

Yes, please.

> I suspect that the bundle problem in travis here will repeat as I did not do anything against it. Any ideas please ?

There's nothing to be done here. This is waiting on https://github.com/openmicroscopy/ansible-role-jekyll-build/pull/10

sbesson:
#37 consumes the changes proposed openmicroscopy/ansible-role-jekyll-build#10 and seems to suffice to fix the Travis build. I suggest we give it a proper review tomorrow and determine the best strategy between:

1- accepting the upstream role changes in order to fix the Travis build for prod-playbooks
2- agreeing that more work/design is required and probably re-allowing Travis failures for the `www` playbook to minimize the impact on other PRs

pwalczysko:
@joshmoore added the new PR at https://github.com/openmicroscopy/management_tools/pull/674

training-server.yml
```diff
@@ -193,8 +221,19 @@
...
       omero.jvmcfg.percent.indexer: 20
       omero.jvmcfg.percent.pixeldata: 20
+      omero.ldap.config: "true"
+      omero.ldap.urls: "ldap://localhost:10389"
+      omero.ldap.base: "dc=openmicroscopy,dc=org"
+      omero.ldap.group_filter: "(objectClass=groupOfUniqueNames)"
+      omero.ldap.group_mapping: "name=cn"
+      omero.ldap.new_user_group: "MyData"
+      omero.ldap.new_user_group_owner: "(owner=@{dn})"
+      omero.ldap.password: "{{ ldap_password | default(secret)}}"
```
manics: `'secret'` needs to be quoted, otherwise it's treated as a variable name

joshmoore: Thanks, @manics. I updated https://github.com/openmicroscopy/prod-playbooks/pull/41#issuecomment-365182258

pwalczysko:
Implemented comments from here and https://github.com/openmicroscopy/management_tools/pull/674

joshmoore:
Sorry for the nitpick, but I think the `apache_docker_release` variable should be defined in this repository so that the playbook is always testable.

pwalczysko:
>apache_docker_release variable should be defined in this repository

I agree, but

1. do not see an obvious place to put it (no ``host_vars``)
2. the https://github.com/openmicroscopy/prod-playbooks/pull/38 which defines analogical variables does not mention their definitions at all

Is there any implicit system for this which I am missing ?

joshmoore:
> do not see an obvious place to put it (no host_vars)

In the file itself, you can add a `vars:` section. Please see https://github.com/openmicroscopy/prod-playbooks/blob/master/ome-dundeeomero.yml#L252

> the #38 which defines analogical variables does not mention their definitions at all

In the cases defined by @manics, the value can easily be left blank. Simon's using `|default(omit)` which means "if no value is set, then act as if this key/value pair did not exist in the file".

pwalczysko:
>In the file itself, you can add a vars: section.

Yes, so supposing a run from MT repo, the var will be defined twice, once in the playbook itself, the other time in the ``outreach....yml`` file in ``host_vars`` folder of the MT themselves. Which value will prevail (or will it all crash ?) ? It seems that my mental image of the workflow(s) of running the playbook is not precise enough... I actually found it a bit surprising that the playbook can be run from this repo directly at all.

joshmoore:
> Which value will prevail (or will it all crash ?) ? It seems that my mental image of the workflow(s) of running the playbook is not precise enough...

@pwalczysko : see http://docs.ansible.com/ansible/latest/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

> In 2.x, we have made the order of precedence more specific (with the last listed variables winning prioritization):
>
> * role defaults [1]
> * inventory file or script group vars [2]
> * inventory group_vars/all
>  * playbook group_vars/all
>  * inventory group_vars/*
>  * playbook group_vars/*
>  * inventory file or script host vars [2]
>  * inventory host_vars/*
>  * playbook host_vars/*
>  * host facts
>  * play vars
>  * play vars_prompt
>  * play vars_files
>  * role vars (defined in role/vars/main.yml)
>  * block vars (only for tasks in block)
>  * task vars (only for the task)
>  * role (and include_role) params
>  * include params
>  * include_vars
>  * set_facts / registered vars
>  * extra vars (always win precedence)


I realize this is a bit daunting, but the power of ansible variables is eventually worth it. Simplifying this list a bit for what you are trying to do:

 * **inventory host_vars** - the file in MT
 * *is overwritten by...* **play vars** - the `vars:` block in your `training-server.yml`
 * *is overwritten by...* **task vars** -- if you also define something under a given task under `tasks:`
 * *is overwritten by...* **extra vars** -- if you pass something on the command-line (e.g. `-e apache_docker_version=0.4.0`)


>  I actually found it a bit surprising that the playbook can be run from this repo directly at all.

Let's deal with this next

joshmoore:
Sorry, @pwalczysko, I think I was unclear. Since the values in this file have a higher precedence than (i.e. will overwrite) those in the host_vars file, this won't quite work yet. Noticed that in the strategy outlined in https://github.com/openmicroscopy/prod-playbooks/pull/41#issuecomment-365182258 , the value coming from the host file are different from those used in this file. A way to think about this is likely to have a naming scheme for the variables that are in the host file to make it clear that they are _extra_ variables which are to be used as overrides:

```
# in the playbook

  vars:
    ...
    os_system_users_password: "{{ os_systems_users_password_override | default('$6$leKi5...oZPMqGzN/')}}"
    apache_docker_release: "{{ apache_docker_release_override | default('0.3.0') }} "
    etc.
```

If you *don't* think that you'll need an override and the value is not *secret*, then it's probably simpler to specify it directly in the playbook as you have here:

```
    apache_docker_release: 0.3.0
```

pwalczysko:
@joshmoore : Okay, I think I have it working as per your last comment locally. Upgraded to the ``0.4.0`` version of ``apacheds-docker``. I can verify that the docker itself as installed on outreach is ``0.4.0`` (at least it is named so in the ``docker ps`` output). But 

How would I verify the version of the installed ``ldapmanager`` script please ?

joshmoore:
There's no way other than comparing the content. We would need to migrate `ldapmanager` to use https://github.com/openmicroscopy/yaclifw and then likely push to pypi, etc.

joshmoore:
Thanks for helping to discover and refine this logic, @pwalczysko !

joshmoore:
:+1: for the changes. Running locally I see the now common

```
$ ansible-playbook -i ../ansible/inventory/prod-hosts playbooks/training-server.yml --ask-become -C
...
outreach.openmicroscopy.org : ok=100  changed=12   unreachable=0    failed=0
```

At least in check-mode, it looks like the script would be downloaded. Otherwise, :+1:

PR #43, opened 2018-02-28
Playbook naming unification

sbesson:
As discussed recently this PR starts unifying the name/layout of some of the web playbooks in this repository to comply with the proposed structure:

```
folder/{folder2}/
   playbook.yml
   molecule.yml
   sub_playbook_1.yml
   ...
   tests/
```

- the OME Website main playbook is renamed as `playbook.yml` and all references are updated
- the syntax-check `molecule.yml` is copied under `web-proxy/molecule.yml`
- the `web-proxy`  check (syntax only) is removed from the allowed failures in Travis

Functionally this change should have no impact as shown by the Travis builds.


web-proxy/molecule.yml
```diff
@@ -0,0 +1,26 @@
...
+dependency:
+  name: galaxy
+  requirements_file: ../requirements.yml
+
+driver:
+  name: docker
+
+docker:
+  containers:
+  - name: host-populate
```
manics: I don't think this matters since it's only a syntax check, but this should ideally match the hosts in the playbook: `prod-web-proxies`

PR #45, opened 2018-03-02
www: fix playbook execution with Ansible 2.4.3.0

sbesson:
Ansible 2.4.3.0 has a known issue when git clone using HEAD as a version. See https://github.com/ansible/ansible/issues/30780 

This PR allows to deploy the OME website using Ansible 2.4 by bumping the `openmicroscopy.jekyll-build` role to version 1.3.1 which sets a default value to the Git branch.

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -3,8 +3,7 @@
 - src: rvm_io.ruby
   version: v2.0.1
 
-- name: openmicroscopy.jekyll-build
-  src: https://github.com/sbesson/ansible-role-jekyll-build/archive/default_branch.tar.gz
+- src: openmicroscopy.jekyll-build
   version: 1.3.1
 
 - src: openmicroscopy.lvm-partition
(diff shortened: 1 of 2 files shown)


joshmoore:
If this is going to be the default behavior from https://github.com/ansible/ansible/commit/99ef1f3a9fa2bc16f47ce97c353c66da2efa8336 I wonder if (additionally) we need to drop the `|default(omit)` logic and just hard-code `master` as the default in https://github.com/openmicroscopy/ansible-role-jekyll-build/blob/master/defaults/main.yml#L14

sbesson:
Opened as https://github.com/openmicroscopy/ansible-role-jekyll-build/pull/11, should I update this PR to use `requirements.yml` in addition or as replacement of e755715?

sbesson:
0ca6420  updates the requirements file to consume the patch release of `openmicroscopy.jekyll-build`

joshmoore:
LGTM. Thanks, @sbesson

PR #46, opened 2018-03-05
Fix artwork redirect

hflynn:
Previous implementation isn't working for https://www.openmicroscopy.org/site/support/ome-artwork/artwork-usage

### DIFF
--- a/www/www-deploy.yml
+++ b/www/www-deploy.yml
@@ -137,7 +137,7 @@
     # support
     - match: "~/site/support/?$"
       dest: /docs
-    - match: "~/site/support/ome-artwork/?$"
+    - match: "~/site/support/ome-artwork/(?<link>.*)$"
       dest: /artwork
     - match: "~/site/support/(?<link>.*)$"
       dest: https://www-legacy.openmicroscopy.org/site/support/$link
--- a/www/www-deploy.yml
+++ b/www/www-deploy.yml
@@ -137,7 +137,7 @@
     # support
     - match: "~/site/support/?$"
       dest: /docs
-    - match: "~/site/support/ome-artwork/.*$"
+    - match: "~/site/support/ome-artwork(/.*)?$"
       dest: /artwork
     - match: "~/site/support/(?<link>.*)$"
       dest: https://www-legacy.openmicroscopy.org/site/support/$link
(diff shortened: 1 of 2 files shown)


www/www-deploy.yml
```diff
@@ -137,7 +137,7 @@
     # support
     - match: "~/site/support/?$"
       dest: /docs
-    - match: "~/site/support/ome-artwork/?$"
+    - match: "~/site/support/ome-artwork/(?<link>.*)$"
```
sbesson: This can probably be reduced to `~/site/support/ome-artwork/.*$` (to check with regexp) since we are not using the last part of the URL in the destination link

hflynn: Done

hflynn:
Hopefully sorted now

www/www-deploy.yml
```diff
@@ -137,7 +137,7 @@
     # support
     - match: "~/site/support/?$"
       dest: /docs
-    - match: "~/site/support/ome-artwork/?$"
+    - match: "~/site/support/ome-artwork/.*$"
```
sbesson: Actually with this change, the form with no trailing slash i.e. '/site/support/ome-artwork' will not match and be redirected. We either need two redirects (`/?$` and `/.*$`) or to find a regexp way to combine them to capture all use cases

hflynn: Shall I add two for now?

sbesson: Unless you have the syntax for combining all use cases in one (i don't know it off-hand), two might be the fastest for now.

kennethgillen: I'm not sure why the rest of the forward slashes aren't escaped, but I'd try:

`"~\/site\/support\/ome-artwork(\/.*)?$"`

This says "ome-artwork" on it's own, or followed by at least `/` then zero-or-more of any other character.

i.e. the `/` is required if there will be any more characters after `ome-artwork`

kennethgillen: see https://regexr.com/3lpp9 cc @hflynn @sbesson

kennethgillen: i.e. you can leave out the `\` before the forward slashes if you like, as if it's working for the rest then it'll work for this: but generally one would escape them. (That does depend on the flavor of Regex mind)

hflynn:
Hopefully sorted now, thanks @kennethgillen

sbesson:
Deployed on https://www.openmicroscopy.org/

PR #47, opened 2018-03-05
Add redirect for script sharing page

hflynn:
I thought this was already in place but it seems not.

Redirects the scripts page to the docs page which has been updated to more closely match the content.

This leaves only the minutes and meeting item of the community menu still going to the legacy URL

### DIFF
--- a/www/tests/check_redirects.py
+++ b/www/tests/check_redirects.py
@@ -56,7 +56,6 @@ external_uris = [
     ('/site/community/scripts', 'https://docs.openmicroscopy.org/latest/omero/developers/scripts/index.html'),
 ]
 legacy_uris = [
-    '/site/community/scripts',
     '/site/community/minutes',
     '/site/community/minutes/conference-calls/2017',
     '/site/products/ome-tiff',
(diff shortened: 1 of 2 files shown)


www/tests/check_redirects.py
```diff
@@ -53,6 +53,7 @@
 external_uris = [
     ('/omero-blog', 'http://blog.openmicroscopy.org'),
     ('/site/about/development-teams/glencoe-software', 'https://www.glencoesoftware.com/team.html'),
+    ('/site/community/scripts', 'https://docs.openmicroscopy.org/latest/omero/developers/scripts/index.html'),
 ]
 legacy_uris = [
     '/site/community/scripts',
```
sbesson: This should probably be removed in favor of the test above?

hflynn:
@sbesson done - sorry I hadn't spotted that list

sbesson:
Looks good now. Will retest tomorrow with the other changes from #46.

sbesson:
Deployed on https://www.openmicroscopy.org/

PR #51, opened 2018-03-16
www playbook: consume new openmicroscopy.ssl-certificate role

sbesson:
See https://github.com/openmicroscopy/ansible-role-ssl-certificate/pull/1 and companion IDR PR https://github.com/IDR/deployment/pull/97

This PR makes the required adjustements to the `www` playbook to consume the proposed new role.

## Changes

- set `jekyll_build_force_rebuild` to `False` and re-include molecule idempotence tests in Travis
- update `requirements.yml` to consume proposed `openmicroscopy.self-certificate` role
- update `www-deploy.yml` to consume the new role with the corresponding variable changes (no more pre tasks should be required).

## Testing

- Travis should pass. The `www` component of the matrix should run the full `molecule test` sequence include the idempotence and  test the self-signed certificate
- the playbook should be run against the `ome-www-dev` server


sbesson:
Latest version of this PR with the new Galaxy role now deployed against `ome-www-dev`. Leaving @kennethgillen and/or @joshmoore to give a final round of code review, then I propose we deploy it on the production www and create an issue for propagating this new role across the other playbooks.

www/www-deploy.yml
```diff
@@ -2,37 +2,26 @@
...
-      nginx_proxy_ssl_certificate_key: "{{ nginx_ssl_files_path }}/{{ nginx_ssl_key_filename }}"
-      nginx_proxy_http2: True
-      nginx_proxy_force_ssl: False
-      nginx_proxy_404: "/404.html"
-      nginx_proxy_conf_http:
-        - "client_max_body_size 2g"
 
   vars:
-
+    ssl_certificate_public_path: /etc/nginx/ssl/server.crt
```
joshmoore: These could have also been left as the default, no?

sbesson: I used the application-specific SSL directory to minimize confusion after discussion with @kennethgillen. Using the default should equally work.
Since the variable is coupled to `nginx_proxy_ssl_certificate` below, my impression is that explicitly declaring its value prevents issues if the defaults were to change.

joshmoore: Ah, I just found the discussion around https://serverfault.com/a/840191 -- if this is the reasoning, should be not adjust the upstream role to something like `/etc/ssl/local_certs`?

kennethgillen: @joshmoore - `/etc/ssl/local_certs` could work, to be application-agnostic.

sbesson: See https://github.com/openmicroscopy/ansible-role-ssl-certificate/pull/5 for the proposed role change.

sbesson:
With the last commit, this was successfully run against `ome-ww-dev`. Proposing to get this merged and deploy to the production server alongside https://github.com/IDR/deployment/pull/97#issuecomment-382009005

PR #52, opened 2018-03-19
Remove all legacy redirects, add events, info & fix licensing

hflynn:
As discussed at today's web meeting, this removes all the redirects to www-legacy (see https://trello.com/c/7GTUHoJF/143-timeline-for-turning-off-legacy-and-redirecting-everything)

I've also added an events redirect as suggested on https://trello.com/c/nsjt36jH/90-broken-omero-2017-user-meeting as i forgot about those URLs when I was adding the meetings redirects.


hflynn:
Is Travis broken because my test addition is wrong or is it unrelated to this PR?

joshmoore:
Unrelated (nightshade download timed out):

```
 [WARNING]: - openmicroscopy.postgresql-3 was NOT installed successfully:
Failed to get data from the API server (https://galaxy.ansible.com/api/): ('The
read operation timed out',)
```

I've restarted it.

hflynn:
Thanks!

hflynn:
2 new commits to address https://trello.com/c/riJnvMrD/147-downloads-page-cc-license-link

sbesson:
NB: this was incidentally deployed to `ome-www-dev` as part of https://github.com/openmicroscopy/prod-playbooks/pull/51#issuecomment-375746661 for anyone who wants to review.

www/www-deploy.yml
```diff
@@ -69,8 +69,6 @@
       dest: /security/advisories/
     - match: "~/info/vulnerabilities/(?<link>.*[^/])/?$"
       dest: /security/advisories/$link/
-    - match: "~/info/(?<link>.*)$"
```
joshmoore: What now handles the `/info/` URLs?

www/www-deploy.yml
```diff
@@ -121,14 +119,14 @@
       dest: /bio-formats/downloads/
     - match: "~/site/products/ome-files-cpp/?$"
       dest: /ome-files
-    - match: "~/site/products/(?<link>.*)$"
```
joshmoore: Similarly, how is this handled now? Also, why is `ome-files-cpp` still above?

www/www-deploy.yml
hflynn: They'll just go the front page of the new website, won't they? I don't know what they are to redirect them individually

www/www-deploy.yml
hflynn: I can redirect it to /products if you like but this covers the partner products we don't have pages for any more.

Why would ome-files-cpp not be here? We have a specific page to redirect that to.

www/www-deploy.yml
hflynn: Or actually to our new 404 page I guess. I could just redirect them to the home page if you prefer?

hflynn: Unless, I guess there must be a list somewhere because the legacy/info/ URLs are all redirects anyway - any idea where they live @sbesson ?

sbesson: I do not think dropping them is an option. A [search](https://github.com/search?q=%22openmicroscopy.org%2Finfo%22&type=Code&utf8=%E2%9C%93) on GitHub only shows minimally 1K usages and some of those `info` links are used for the security vulnerability pages or attribution links in the core xsd schema files.

The list is defined in the Apache configuration file for `www-legacy`. Reducing the `www/info -> www-legacy/info -> link` redirection path might be valuable. Happy to have a call about this tomorrow but I would propose we handle it outside the scope of this PR.

hflynn: The security ones are handled separately above anyway but yes, we can discuss tomorrow

hflynn:
This should address @joshmoore 's concerns, will re-list for Monday

www/www-deploy.yml
```diff
@@ -141,16 +135,58 @@
...
       dest: /artwork
-    - match: "~/site/support/(?<link>.*)$"
-      dest: https://www-legacy.openmicroscopy.org/site/support/$link
+
+    # info
+    - match: "~/info/vulnerabilities/?$"
+      dest: /security/advisories/
+    - match: "~/info/vulnerabilities/(?<link>.*[^/])/?$"
+      dest: /security/advisories/$link/
+    - match: "~/info/omero$"
```
sbesson: The trailing `omero$` fails the test when `/info/omero/` is passed as the URL suffix. Since we are matching regexp, could be turned into `~/info/omero/?$`

sbesson: Also should we handle the capitalized version as well (to match the info redirects below)?

www/www-deploy.yml
```diff
@@ -141,16 +135,58 @@
...
+      dest: /omero
+    - match: "~/info/OMERO.insight$"
```
sbesson: same issue as above (URL with trailing slash)

www/www-deploy.yml
```diff
@@ -141,16 +135,58 @@
...
+      dest: https://docs.openmicroscopy.org/latest/omero/users/index.html
+    - match: "~/info/OMERO.importer$"
```
sbesson: same issue as above

www/www-deploy.yml
```diff
@@ -141,16 +135,58 @@
...
+      dest: https://docs.openmicroscopy.org/latest/omero/users/index.html
+    - match: "~/info/OMERO.editor$"
```
sbesson: same issue as above

www/www-deploy.yml
```diff
@@ -141,16 +135,58 @@
...
+      dest: https://docs.openmicroscopy.org/latest/omero/users/index.html
+    - match: "~/info/OMERO.web$"
```
sbesson: same issue as above

www/tests/check_redirects.py
```diff
@@ -54,11 +56,30 @@
...
+    ('/info/demo', 'http://help.openmicroscopy.org/demo-server.html'),
+    ('/info/lists', 'http://lists.openmicroscopy.org.uk/mailman/listinfo/'),
+    ('/info/videos', 'https://www.youtube.com/channel/UCyySB9ZzNi8aBGYqcxSrauQ'),
+    ('/info/downgrade', 'https://docs.openmicroscopy.org/latest/omero/developers/Model/XsltTransformations.html'),
+    ('/info/flimfit', 'http://flimfit.org'),
+    ('/info/scripts', 'https://docs.openmicroscopy.org/latest/omero/developers/scripts/index.html'),
+    ('/info/bio-formats', 'https://docs.openmicroscopy.org/latest/bio-formats/'),
+    ('/info/slidebook', 'https://www.intelligent-imaging.com/technical-answers'),
 ]
 legacy_uris = [
```
sbesson: These tests are now failing as per the removal of the legacy links.

www/www-deploy.yml
hflynn: OMERO capitalized but without a client or server added to it isn't on the list of info URLs we've ver given out - I didn't want to use this PR to add new redirects

hflynn:
@sbesson hopefully fixed now

sbesson:
```
_______________________________________ ERROR collecting playbooks/www/tests/check_redirects.py _______________________________________
playbooks/www/tests/check_redirects.py:139: in <module>
    @pytest.mark.parametrize('uri', legacy_uris)
E   NameError: name 'legacy_uris' is not defined
```

Can you also remove the test itself?

sbesson:
Apart from https://github.com/openmicroscopy/prod-playbooks/pull/52#issuecomment-378210379, this deployed correctly on `ome-www-dev` and all redirects tests are passing.

The only remaining question is whether redirecting some of the removed wildcard regexp to some known pages (like `/info/.*` -> `/site-map`) rather than using the 404 page would provide a better user experience. Looking at the details, my two candidates open to debate: would be the following

```
/site/support/.* -> /docs
/site/community/.* -> /support
```

hflynn:
The other /support/ URLs are for partner project docs - there's nothing on /docs that's equivalent and arguably users are more likely to find the content they're looking for via the OMERO features sections as that's where the links for FLIMfit and webtagging for example, live now. So I don't think redirecting them to /docs is helpful.

As for other /community/ links - they're a bit random but I can certainly see that /support might be helpful so I could add that.

sbesson:
Last commits look good. I'll deploy them for a final round of review tomorrow.

sbesson:
Redeployed on `ome-www-dev`. All mapping added above redirect to the corresponding pages. All updated redirect tests are also passing.

sbesson:
Now deployed live on https://www.openmicroscopy.org/

hflynn:
Spotted one more construction of the licensing link on downloads which is still pointing to the top level About page - I'll open a new PR to fix that.

PR #58, opened 2018-04-16
Add remaining redirects from plone

hflynn:
Implements the final redirects we want to keep from the old plone web-prod config - see https://docs.google.com/spreadsheets/d/12Tsmyg3r7kOSDYiupKW2dA_5wTltlecSHd3TzUAdJHc/edit#gid=0 (with the exception of the BF jars latest downloads redirects which should live elsewhere)

Once this PR is finalised and merged, I assume we can just delete https://github.com/openmicroscopy/management_tools/blob/master/prod/config/ome-web-prod/httpd/plone-redirects.include entirely or do we need a skeleton file left?


hflynn:
Will add Bio-Formats jars

hflynn:
Actually, there are no downloads redirects on this page, I assume they live somewhere else

sbesson:
Proposed redirects work as expected. Merging for deployment on the live website. Regarding the remaining cleanup, I need to review the state but I think :
- we should almost be ready to invert the paradigm and implement a general `www-legacy.openmicroscopy.org/site` -> `www.openmicroscopy.org/site/*` redirect
- we probably still need at least one redirect PR to handle the remaining top-level non `/site` URLs (like `/formats`)

PR #59, opened 2018-04-16
Systemd dependency for ome-dundeeomero

kennethgillen:
Fixes the startup of OMERO on ome-dundeeomero which requires GPFS to be loaded and running before OMERO will properly start.

At the moment, a reboot of the server (normal during patching) will result in an OMERO instance running with an inaccessible OMERO.datadir


joshmoore:
Restarted travis after "too many 503s"

joshmoore:
Travis now green. Additionally, having released 1.0.3 & 2.0.5 yesterday, this change makes sense to me.

Issue #24, opened 2017-12-18
Define use of tags


joshmoore:
See discussion under https://github.com/openmicroscopy/prod-playbooks/pull/14#discussion_r157182379

manics:
If tags are at the level of tasks in roles (which in #14 would make sense) this needs to managed at the organisation level, since:
- If someone is running a playbook they shouldn't need to read the docs for all roles in the playbook to check what tags are used and their effect
- Tags need to be consistent in term of what the do across all roles/tasks/playbooks/repos we use
- They become public and exposed to external users of our roles
- If we decide to incorporate roles/playbooks from other sources they'll need to be reviewed to ensure there are no conflicting tags

joshmoore:
An aside I've mentioned in passing before:

 * tags on role inclusion may be an easy first step here.

See:

 * https://ansiblemaster.wordpress.com/2017/01/18/using-tags-for-each-role/

```
 roles:
    - { role: frontend1,
        tags: frontend1 }
    - { role: frontend2, 
        tags: frontend2 }
```

kennethgillen:
Current tag set:

```
2018-04-16 16:48 LS27172:prod-playbooks kenny$ git grep tags | grep -Po '(?<=\btags\: )[^\s]+' | sort | uniq -c | sort -rn
  18 monitoring
  14 lvm
   2 backups
   1 redirect
   1 nginxconf
   1 jekyll
```

kennethgillen:
Would @openmicroscopy/devops object to tag `network`?

I've added it to --skip-tags on my local test deployment of `web-proxy`, which effectively makes the web-proxy playbook like `www` https://github.com/openmicroscopy/prod-playbooks/tree/master/k8s where there are bootstrap steps, then config steps which can be reapplied.

There's the same concept for the CI 1.5 servers in this folder - i.e. the `bootstrap` folder https://github.com/openmicroscopy/prod-playbooks/tree/master/bootstrap

Perhaps all of `lvm`, `network`, .. whatever else, should be tagged `bootstrap`, too? 

This concept basically encompasses the issue I had with the ome-dundeeomero playbook back at the start: i.e. there are things commented out which were necessary for the first run, but not subsequent.

Having a folder-per-playbook, with a subfolder called `bootstrap` would be one way to deal with this, which we decided not to do I think at the start, even though `www` seems to have gone that way.

joshmoore:
[...]

PR #60, opened 2018-04-17
Adding 'bootstrap' task tags for testing

kennethgillen:
Adding tags to allow `--skip-tags` when executing, e.g. a local test pipeline which will be adversely affected by such changes as storage and networking, which differ to the `prod` config.

See https://github.com/openmicroscopy/prod-playbooks/issues/24#issuecomment-381656949 for more rationale.

### DIFF
--- a/web-proxy/playbook.yml
+++ b/web-proxy/playbook.yml
@@ -4,12 +4,15 @@
 - hosts: prod-web-proxies
   roles:
   - role: openmicroscopy.network
+    tags: network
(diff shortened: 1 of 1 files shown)


joshmoore:
These tasks look inline with what we currently have in place. As mentioned on gh-24, we'll have to come back and review and likely _change_ these so though they are convenient for our own use, we should review them as public API.

kennethgillen:
Thank you.

PR #53, opened 2018-03-20
new server pub-omero.openmicroscopy.org

kennethgillen:
Provision of new server pub-omero.openmicroscopy.org
Relevant MT PR: https://github.com/openmicroscopy/management_tools/pull/730

Rationale for renaming `vault` -> `secrets`: https://github.com/openmicroscopy/prod-playbooks/pull/53#issuecomment-375348170

PR #62, opened 2018-04-25
pub-omero equivalent to demoserver

kennethgillen:
Fixed up https://github.com/openmicroscopy/prod-playbooks/pull/53 which had a some revert commits and a rebese gone badly wrong.

### DIFF
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -10,7 +10,7 @@
 
 # To allow the OMERO.web plugins to upgrade, also pass `--extra-vars upgrade_webapps=True`
 
-- hosts: ome-demoserver.openmicroscopy.org
+- hosts: ome-demoservers
   pre_tasks:
     - name: Install open-vm-tools if system is a VMware vm
       become: yes
@@ -119,7 +119,7 @@
         databases: ["{{ vault.omero_server_dbname }}"]
 
     - role: openmicroscopy.omero-server
-      omero_server_release: 5.4.5
+      omero_server_release: "{{ omero_server_release_desired | default('5.4.5') }}"
       omero_server_dbuser: "{{ vault.omero_server_db_user }}"
       omero_server_dbname:  "{{ vault.omero_server_dbname }}"
       omero_server_dbpassword: "{{ vault.omero_server_dbpassword }}"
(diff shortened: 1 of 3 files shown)


kennethgillen:
cc @sbesson @jburel

kennethgillen:
`--check` mode run: http://beluga.openmicroscopy.org:8000/reports/d2a93600-43e4-474f-a978-4b3c925f5ae7.html

ome-demoserver.yml
```diff
@@ -119,15 +119,15 @@
...
+      omero_server_release: "{{ omero_server_release_desired | default('5.4.5') }}"
       omero_server_dbuser: "{{ vault.omero_server_db_user }}"
       omero_server_dbname:  "{{ vault.omero_server_dbname }}"
       omero_server_dbpassword: "{{ vault.omero_server_dbpassword }}"
       omero_server_rootpassword: "{{ vault.omero_server_rootpassword }}"
       omero_server_systemd_limit_nofile: 16384
 
     - role: openmicroscopy.omero-web
-      omero_web_release: 5.4.5
+      omero_web_release: "{{ omero_server_release_desired | default('5.4.5') }}"
```
sbesson: `s/omero_server_release_desired/omero_web_release_desired`?

kennethgillen: d'oh - thanks. Updated and pushed.

kennethgillen:
I don't particularly like:

```
PLAY [ome-demoservers] **********************************************************************************
 [WARNING]: While constructing a mapping from
/Users/kenny/forks/management_tools/ansible/inventory/group_vars/vlan-10ge-servers.yml, line 2, column
3, found a duplicate dict key (device). Using last defined value only.

```

but am willing to live with it for now, personally, given it does what I want.

requirements.yml
```diff
@@ -10,7 +10,7 @@
   version: 1.1.0
 
 - name: openmicroscopy.network
-  version: 1.1.1
+  version: 1.1.0
```
sbesson: This version number decrement breaks the execution of `web-proxy` playbook. The changes proposed in https://github.com/openmicroscopy/ansible-role-network/pull/3 and released in `1.1.1` were required to fix the playbook execution with the current inventory variables. 

The issue that this tries to address was reported as https://github.com/openmicroscopy/ansible-role-network/issues/4.  Can we follow-up on https://github.com/openmicroscopy/ansible-role-network/issues/4#issuecomment-374883606 to decide whether 1- we can get the role fixed for all usages, 2- and/or there is a need to unify our internal variables definition.

kennethgillen: @sbesson - this change is no longer in the diff on this PR, so this is addressed by https://github.com/openmicroscopy/prod-playbooks/pull/62/commits/fa7838380e95abce056c36a3947bb8d56c0860b4

kennethgillen:
Thanks @sbesson. We should be good to merge then.

PR #63, opened 2018-05-09
Add new playbook to update public OME presentations

sbesson:
/cc @jburel

### DIFF
--- a/release/presentations.yml
+++ b/release/presentations.yml
@@ -1,0 +1,10 @@
+---
+- hosts: idr0-slot3.openmicroscopy.org
+  become: true
+  tasks:
+    - name: update presentations
+      git:
+        repo: https://github.com/ome/presentations
+        force: no
+        update: yes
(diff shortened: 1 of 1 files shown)


sbesson:
Executed in check-mode (see `reports/612a4ac0-6ced-4f92-8ae9-7ced50d6586b.html`).
For all the playbooks creating/adding static content, I am considering adding an extra task to all playbooks outside check-mode to create a GPFS snapshot following their execution.

manics:
Looks fine.

`$ git status --ignored` in the existing checkout shows several video files and pdfs. How are those handled?

sbesson:
All binaries are managed completely manually and this remains a requirement until we come up with a better solution. Just realized while talking to @jburel that this playbook does not make the folders group-writeable (and ideally with sticky bit)

sbesson:
Merging to test the work flow with @jburel

PR #64, opened 2018-05-09
training-server.yml molecule test

manics:
In the next few weeks several people including myself and @joshmoore may make changes to the outreach. This PR does the minimal work necessary to get a basic molecule test running on travis before we start opening PRs:

- Moves `training-server.yml` into it's own subdirectory with **minimal** changes. Cleanups and improvements can be handled in a separate PR.
- Changes `training-server.yml` to use the correct postgres role. This playbook is broken without this, but was probably unnoticed because the original server was deployed from a different repository and this bug is only noticeable on a new deployment.
- Bumps the docker role version in an attempt to work around a docker bug (not sure if this is a docker for mac bug, an `openmicroscopy/apacheds:0.3.0` bug, or something else (I've tested other playbooks and roles that use docker in docker in a similar manner to this playbook and they work fine). This should be safe since the change adds an optional variable: https://github.com/openmicroscopy/ansible-role-docker/compare/2.2.0...2.3.0

Notes:
- The idempotence check is disabled since some tasks don't pass
- I've run this with `molecule test --driver vagrant --destroy never` locally. I've attempted to test with docker locally too but there's some weird bug... let's see what travis thinks.
- There's already a discrepancy between the OMERO version in this playbook (5.4.3) and that on the real server (5.4.4)
- For the molecule test **only** I'm overriding a few variables to get around various bugs, see `ansible.extra_vars` in `molecule.yml`

@pwalczysko

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ env:
   - DIRECTORY="." PLAYBOOK=nightshade-webclients.yml
   - DIRECTORY="." PLAYBOOK=ome-demoserver.yml
   - DIRECTORY="." PLAYBOOK=ome-dundeeomero.yml
-  - DIRECTORY="." PLAYBOOK=training-server.yml
+  - DIRECTORY="training" PLAYBOOK=training-server.yml
   - DIRECTORY=bootstrap PLAYBOOK=
   - DIRECTORY=web-proxy PLAYBOOK=
   - DIRECTORY=www PLAYBOOK=
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ env:
   - DIRECTORY="." PLAYBOOK=nightshade-webclients.yml
   - DIRECTORY="." PLAYBOOK=ome-demoserver.yml
   - DIRECTORY="." PLAYBOOK=ome-dundeeomero.yml
-  - DIRECTORY="training" PLAYBOOK=training-server.yml
+  - DIRECTORY=training PLAYBOOK=
   - DIRECTORY=bootstrap PLAYBOOK=
   - DIRECTORY=web-proxy PLAYBOOK=
   - DIRECTORY=www PLAYBOOK=
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ env:
   - DIRECTORY="." PLAYBOOK=nightshade-webclients.yml
   - DIRECTORY="." PLAYBOOK=ome-demoserver.yml
   - DIRECTORY="." PLAYBOOK=ome-dundeeomero.yml
-  - DIRECTORY=training PLAYBOOK=
+  - DIRECTORY=omero/training-server PLAYBOOK=
   - DIRECTORY=bootstrap PLAYBOOK=
   - DIRECTORY=web-proxy PLAYBOOK=
   - DIRECTORY=www PLAYBOOK=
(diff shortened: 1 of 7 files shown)


pwalczysko:
See snippet out of my presentation for next week Cambridge (which will be most probably re-used by @jburel during User's meeting. We cite just the prod-playbooks repo, not concretely the .yml, so this should be fine with the subfolder planned here.
screen shot 2018-05-10 at 09 55 31

sbesson:
The Travis issue should be fixed once we release OMERO 5.4.6 publicly later today. For the top-level folder, an unanswered proposal from February was to eventually group all OMERO playbooks under a top-level `omero` folder. Any reason to treat `training` differently?

manics:
I must have missed the proposal. How would molecule tests fit into that framework?

pwalczysko:
>Any reason to treat training differently?

I would definitely okay this with @jburel first.

sbesson:
To broaden the context, below is a copy-paste of an email extract from February 19th discussing the organization of these playbooks:

```
C) prod-playbooks

Spinning up a k8s cluster is performed by ansible, so the playbooks
for this are slotted for addition to prod-playbooks [3]. There are
currently a number of different ways that playbooks are laid out
there:

(1) playbooks at the top
(2) playbooks in subdirectories
  - named playbook.yml or
  - named something specific
(3) playbooks in sub-subdirectories

Assuming we don't want all the files flat at the top-level (1), would
this hierarchy make general sense (to those who have any familiarity
with this repo):

k8s/
 bootstrap
 postgres
omero/
 dundee
 demo
 outreach
web/
  proxy/
  site/
```

Since no-one objected either by email or during one of the SA meeting, I was definitely working under the assumption that this layout would work for all of us. As mentioned yesterday, my plan after fixing presentations was about to act on the last group of playbooks i.e. `website/{playbook,web-proxy,www,...}.yml`.

Re molecule, I understand the idea was to each top-level folder defining its own `molecule` framework testing either part or all of the playbooks. Do you see obvious issues with this?

pwalczysko:
>...this layout would work for all of us. 

I guess it will. Note https://github.com/ome/training-repos link to prod-playbooks will not break I guess.

manics:
@sbesson would the training playbook be `omero/outreach.yml`, or `omero/outreach/playbook.yml`? If the latter that's fine. Do you want me to move it now? I won't touch the other omero playbooks since they're also in production.

sbesson:
:+1: for limiting the scope of this to the training server.

- `omero/outreach.yml` vs `omero/outreach/playbook.yml`: I keep getting back and forth between the two versions. The flattened one is certainly nicer to see all playbooks at once. I assume the second one gives the most flexibility for per-application molecule set up i.e.

   ```
   <category>/
     <application>/
        playbook.yml 
        molecule.yml
   ```
   Are there official recommendations on this?
- `omero/outreach` vs `omero/training-server` are equally fine with me. I can see `outreach.oo.org` being an instance of a `training-server` or `outreach` being used both for the playbook and the host.

manics:
With molecule 1.25 if everything is in the same directory then all playbooks and servers have to be tested simultaneously which will probably hit resource limits, so separate directories is a lot easier. In molecule 2 you can setup scenarios so in theory a single directory would work.

@pwalczysko what directory name do you want, `outreach` or `training-server`?

pwalczysko:
I guess ``training-server``. IMHO it is more precise than ``outreach``. We use ``outreach`` mainly for simplicity, as during the trainings it is easier to type it out, one-word, remember, quickly help others etc. In the github repo however, we do not have such constraints.

manics:
How's this?

PR #67, opened 2018-05-18
Update outreach

pwalczysko:
This PR is attempting several things:

   - have a system of versions captured in variables under ``vars`` section of the playbook, thus making those variables public (discussed recently on a call)
   - use those variables for versioning of OMERO, web and webapps as well as cli plugins
   - add new folder for a new server side script as well as install that script which is needed for users meeting (cc @will-moore )
   - accept the ``openmicroscopy.postgresql-3`` as the role to use for postgres in this playbook (although this was done just by checking out ``master`` from this repo, the playbook was never run on outreach yet with this role)
   - abandon running this playbook from MT, instead run it always from this repo, using ``ansible-playbook -i /Users/pwalczysko/Work/management/ansible/inventory/prod-hosts --become  omero/training-server/playbook.yml  --diff`` command
   - adjust the version of ``OMERO.iviewer`` using the ``test.pypi`` routine, because that matches the status quo on the server. The plan is to use rather ``Pypi``, but the release (0.5.x) will happen only next week. Thus the introducing of this routine makes sure that the playbook will not accidentally downgrade the iviewer to the last released version (0.4.x)

To test:

- The playbook was already tested on ``outreach``, with success. The ``outreach`` is now on 5.4.6 OMERO and web, as well as ``render`` cli plugin is on ``0.3.0``. The ``test.pypi`` routine caused that ``iviewer`` is on ``1.0.1`` which is a ``test.pypi`` version only, and thus there are no changes as compared to the state in the TIM outreach workshop, as intended.
- Check the sensibility of the changes, especially the version vars. After consultation with @kennethgillen , I actually now govern the versions via the ``defaults`` of the version vars in the ``vars:`` section of the playbook @joshmoore @sbesson . This kind of makes sense to me, but open for suggestions. There definitely might be easier systems.

cc @jburel With this playbook, the further upgrades of ``outreach`` should be just matter of re-running the playbook with changes of the version vars (except for iviewer, where we will abandon the ``test.pypi`` routine in favor of simple pypi install).

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -142,7 +142,7 @@
     - name: Download the batch_roi_export_to_table.py script
       become: yes
       get_url:
-        url: https://raw.githubusercontent.com/will-moore/training-scripts/b398d7526e48df79a6c33ce61a679d1458a15ded/practical/python/server/batch_roi_export_to_table.py
+        url: https://raw.githubusercontent.com/ome/training-scripts/master/practical/python/server/batch_roi_export_to_table.py
         dest: /opt/omero/server/OMERO.server/lib/scripts/omero/users_meeting/batch_roi_export_to_table.py
         mode: 0755
         owner: "omero-server"
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -114,11 +114,31 @@
         owner: "omero-server"
         group: "omero-server"
 
+    - name: Create a users_meeting (scripts) directory
+      become: yes
+      file:
+        path: /opt/omero/server/OMERO.server/lib/scripts/omero/users_meeting
+        state: directory
+        mode: 0755
+        recurse: yes
+        owner: "omero-server"
+        group: "omero-server"
+
     - name: Download the Scipy_Gaussian_Filter.py script
       become: yes
       get_url:
         url: https://raw.githubusercontent.com/jburel/omero-example-scripts/cambridge-training/cambridge-training-2017-12/Scipy_Gaussian_Filter.py
         dest: /opt/omero/server/OMERO.server/lib/scripts/omero/cambridge_scripts/Scipy_Gaussian_Filter.py
+        mode: 0755
+        owner: "omero-server"
+        group: "omero-server"
+        force: yes
+
+    - name: Download the Scipy_Gaussian_Filter.py script
+      become: yes
+      get_url:
+        url: https://raw.githubusercontent.com/will-moore/training-scripts/b398d7526e48df79a6c33ce61a679d1458a15ded/practical/python/server/batch_roi_export_to_table.py
+        dest: /opt/omero/server/OMERO.server/lib/scripts/omero/users_meeting/batch_roi_export_to_table.py
         mode: 0755
         owner: "omero-server"
         group: "omero-server"
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -63,24 +63,11 @@
         name:
         - "omero-figure=={{ omero_figure_tag }}"
         - "omero-fpbioimage=={{ omero_fpbioimage_tag }}"
+        - "omero-iviewer=={{ omero_iviewer_tag }}"
         - "omero-mapr=={{ omero_mapr }}"
         - "omero-parade=={{ omero_parade }}"
         - "omero-webtagging-autotag=={{ omero_webtagging_autotag }}"
         - "omero-webtagging-tagsearch=={{ omero_webtagging_tagsearch }}"
-        editable: False
-        state: present
-        # variable comes from role openmicroscopy.omero-web
-        virtualenv: "{{ omero_web_basedir }}/venv"
-        virtualenv_site_packages: yes
-      notify:
-        - restart omero-web
-
-    - name: Omero.web plugins testpypi | plugin install via testpypi
-      become: yes
-      pip:
-        extra_args: -i https://testpypi.python.org/pypi
-        name:
-        - "omero-iviewer==1.0.1"
         editable: False
         state: present
         # variable comes from role openmicroscopy.omero-web
@@ -258,8 +245,9 @@
     omero_server_datadir_chown: True
     omero_server_release: "{{ omero_server_release_override | default('5.4.6') }}"
     omero_web_release: "{{ omero_web_release_override | default('5.4.6') }}"
-    omero_figure_tag: "{{ omero_figure_tag_override | default('3.2.1') }}"
+    omero_figure_tag: "{{ omero_figure_tag_override | default('4.0.0') }}"
     omero_fpbioimage_tag: "{{ omero_fpbioimage_tag_override | default('0.2.0') }}"
+    omero_iviewer_tag: "{{ omero_iviewer_tag_override | default('0.5.0') }}"
     omero_mapr: "{{ omero_mapr_override | default('0.2.3') }}"
     omero_parade: "{{ omero_parade_override | default('0.1.0') }}"
     omero_webtagging_autotag: "{{ omero_webtagging_autotag_override | default('3.0.2') }}"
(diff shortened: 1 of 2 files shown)


jburel:
version on testpy 1.0.2 I think
but that will be changed early next

pwalczysko:
@jburel I did not understand what is the iviewer 1.0.2 (and the figure 3.2.2) on testpypi, so I stuck to what is at present on outreach (iviewer 1.0.1)

omero/training-server/playbook.yml
```diff
@@ -217,6 +251,14 @@
     omero_server_system_managedrepo_group: managed_repo_group
     omero_server_datadir_managedrepo_mode: u=rwX,g=srwX,o=rX,+t
     omero_server_datadir_chown: True
+    omero_server_release_vars: "{{ omero_server_release_override | default('5.4.6') }}"
```
manics: The `_vars` suffix on some of these variable names isn't needed. Could you remove it?

manics:
Travis is failing:
```
TASK [Omero.web plugins testpypi | plugin install via testpypi] ****************
fatal: [outreach.openmicroscopy.org]: FAILED! => {"changed": false, "cmd": "/opt/omero/web/venv/bin/pip2 install -i https://testpypi.python.org/pypi omero-iviewer==1.0.1", "failed": true, "msg": "stdout: Collecting omero-iviewer==1.0.1\n\n:stderr:   Could not find a version that satisfies the requirement omero-iviewer==1.0.1 (from versions: 1.0.2)\nNo matching distribution found for omero-iviewer==1.0.1\nYou are using pip version 9.0.1, however version 10.0.1 is available.\nYou should consider upgrading via the 'pip install --upgrade pip' command.\n"}
```

pwalczysko:
@manics Thank you for clarifying the Travis. This is because there is no version of iviewer 1.0.1 anymore on tespypi. Instead, there is a version called 1.0.2. 

But on outreach server itself, there is 1.0.1 installed, and thus ansible skips any action on that point (that is how i explain to myself why it is not failing on outreach). 
To be fair, I much prefer not trying to install the 1.0.2 version from testpypi because

  - it is one more step which does not add anything valuable for outreach
  - testpypi logic will be abandoned completely
  - this week a pypi version of iveiewer will be released and the normal pypi routine will be introduced in this playbook here as well

What is the best solution in this kind of situation ? I was contemplating completely removing any mentions of iviewer from the playbook for now, but does that guarantee ``no-action`` regarding iviewer when the playbook is playied ?

omero/training-server/playbook.yml
```diff
@@ -63,11 +63,26 @@
...
+        # variable comes from role openmicroscopy.omero-web
+        virtualenv: "{{ omero_web_basedir }}/venv"
+        virtualenv_site_packages: yes
+      notify:
+        - restart omero-web
+
+    - name: Omero.web plugins testpypi | plugin install via testpypi
+      become: yes
+      pip:
+        extra_args: -i https://testpypi.python.org/pypi
```
sbesson: I might have missed a discussion here but why are we installing a version of OMERO.iviewer from test Pypi?

omero/training-server/playbook.yml
```diff
@@ -124,6 +150,16 @@
         group: "omero-server"
         force: yes
 
+    - name: Download the batch_roi_export_to_table.py script
+      become: yes
+      get_url:
+        url: https://raw.githubusercontent.com/will-moore/training-scripts/b398d7526e48df79a6c33ce61a679d1458a15ded/practical/python/server/batch_roi_export_to_table.py
```
sbesson: I assume this URL will be updated to point to the organizational repository? Also does this effectively define a dependency between prod-playbooks and training-scripts down the line i.e. will more training-server scripts be included?

jburel: It should be https://raw.githubusercontent.com/ome/training-scripts/master/practical/python/server/batch_roi_export_to_table.py
now that the PR has been merged

omero/training-server/playbook.yml
jburel: It was before when an rc was tested.
testpypi will no longer be used as discussed last week (you were present)

sbesson: Thanks for the clarification, this is what I remembered. Generally, :+1: for only consuming PyPI packages in our production playbooks. In the cases where we need to use milestones or rcs, we should be able to create [pre-releases](https://www.python.org/dev/peps/pep-0440/#pre-releases) as necessary.

omero/training-server/playbook.yml
```diff
@@ -255,6 +257,8 @@
     omero_web_release: "{{ omero_web_release_override | default('5.4.6') }}"
     omero_figure_tag: "{{ omero_figure_tag_override | default('3.2.1') }}"
     omero_fpbioimage_tag: "{{ omero_fpbioimage_tag_override | default('0.2.0') }}"
+    omero_mapr: "{{ omero_mapr_override | default('0.2.3') }}"
```
jburel: to be consistent it should be ``omero_mapr_tag`` and ``omero_parade_tag``

will-moore:
mapr will need config from https://docs.google.com/document/d/1qXFzeIlijNUDxjRSAIEYZmmeiS4OCREyP_xq7BXOMBs/edit#heading=h.gbtubsz7ki5k
to make it useable.

manics:
Initial monitoring has been added to the outreach server in https://github.com/pwalczysko/prod-playbooks/pull/2

jburel:
Configuration of parade needs to be adjusted.
We should not have a link in the header

omero/training-server/playbook.yml
sbesson: maybe even use `_release` everywhere to match the semantics defined in the `omero-{server,web}` roles

pwalczysko:
Tested on outreach, all went fine, outreach is fine too.

@manics indicated merging would help him to continue better in his workshop prep

jburel:
The script is now at https://raw.githubusercontent.com/ome/training-scripts/master/practical/python/server/scipy_gaussian_filter.py

jburel:
Merging so @manics can move forward and I will open a PR fixing the script link

joshmoore:
:+1: for merging. My only comment that I have for our future refactoring is that I think we need to centralize/rationalize the mapr configuration. All of us have some version of mapr config cut-n-paste (mine is in https://github.com/openmicroscopy/omero-test-infra/pull/12/files#diff-be636689a37ba9fdbb21de2c6616d21aR27) and it seems we could make our lives simpler.

cc: @will-moore

jburel:
and probably parade later on

will-moore:
@joshmoore That links to
```
https://raw.githubusercontent.com/IDR/deployment/{{ idr_deployment_release }}/ansible/group_vars/omero-hosts.yml
```
which I don't understand as mapr config?

joshmoore:
@will-moore : I'm copying the entire IDR config file as my starting point.

will-moore:
OK, so what does that config look like?
Where is the config? I'm looking for something like 
```
{"menu": "gene", "config": {
    "default": ["Gene Symbol"],
    "all": ["Gene Symbol", "Gene Identifier"],
    "ns": ["openmicroscopy.org/mapr/gene"],
    "label": "Gene"}}
```
but all I see is:
```
https://raw.githubusercontent.com/IDR/deployment/{{ idr_deployment_release }}/ansible/group_vars/omero-hosts.yml
```
which means nothing to me.

pwalczysko:
@will-moore 

try https://raw.githubusercontent.com/IDR/deployment/IDR-0.4.5/ansible/group_vars/omero-hosts.yml

you have to resolve the variable idr_deployment_release as  something (I took the value from the playbook @joshmoore linked to). In the link I pasted, search for ``mapr``.

will-moore:
OK, thanks. Sorry, I was just expecting a link direct to the config, not to some code I had to resolve to get a link.
But interesting to see plenty of ```"all": []``` in the config, since I only worked out that this was valid by studying the mapr code.
I wonder if we should make this config public (link to it from https://github.com/ome/omero-mapr.
In fact, I see that I haven't updated that with any examples of ```"all": []```.

PR #69, opened 2018-05-25
omero-web nginx nested includes and letsencrypt

manics:
- Needed so that additional configuration can be added to the generated omero-web nginx config.
- Adds a letsencrypt playbook (automatically disabled in molecule since it requires a matching external DNS record and public http port).


manics:
The nginx changes are blocking the monitoring work.

joshmoore:
Understood. Merging. Can you take responsibility for making sure the refactoring then happens? Thanks.

PR #74, opened 2018-06-13
Revert shared webapp mapr config

manics:
Commit https://github.com/openmicroscopy/prod-playbooks/commit/a5f728b13a0331d5f7030c54d7698fb9a8eb8ab1 from https://github.com/openmicroscopy/prod-playbooks/pull/67 added configuration for mapr and parade to the shared production omero config, however it should only be done on outreach. This PR reverts that commit and replaces it with a file that's only used by outreach.

This suggests we should run all playbooks in this repo with every PR to pick up inadvertent shared changes in future.


kennethgillen:
Well spotted, @manics - thank you.

joshmoore:
see https://github.com/openmicroscopy/prod-playbooks/pull/73#issuecomment-397230421: the assumption is that this will be a no-op. To be confirmed with the re-run from master.

PR #76, opened 2018-06-13
change password

pwalczysko:
Simple change for the outreach playbook.

Consulted with @kennethgillen

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -280,7 +280,7 @@
     omero_webtagging_tagsearch_release: "{{ omero_webtagging_tagsearch_release_override | default('3.0.3') }}"
     omero_cli_duplicate_release: "{{ omero_cli_duplicate_release_override | default('0.2.0') }}"
     omero_cli_render_release: "{{ omero_cli_render_release_override | default('0.3.0') }}"
-    os_system_users_password: "{{ os_systems_users_password_override | default('$6$leKi5B1PgSvdA/ec$xbU3CnoSFnYdeZjEjKK5TH8SGATsW746uopssff4edpgyu.cWXGo9A.oK6wH9kIkxLCCNcORGZnnroZPMqGzN/') }}"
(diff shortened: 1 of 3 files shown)


pwalczysko:
Not sure what is going on with Travis here. It says that there was no output for 10 minutes. The only error which I found was in the ``GPG keys`` paragraph.

omero/training-server/playbook.yml
```diff
@@ -280,7 +292,7 @@
     omero_webtagging_tagsearch_release: "{{ omero_webtagging_tagsearch_release_override | default('3.0.3') }}"
     omero_cli_duplicate_release: "{{ omero_cli_duplicate_release_override | default('0.2.0') }}"
     omero_cli_render_release: "{{ omero_cli_render_release_override | default('0.3.0') }}"
-    os_system_users_password: "{{ os_systems_users_password_override | default('$6$leKi5B1PgSvdA/ec$xbU3CnoSFnYdeZjEjKK5TH8SGATsW746uopssff4edpgyu.cWXGo9A.oK6wH9kIkxLCCNcORGZnnroZPMqGzN/') }}"
+    os_system_users_password: "{{ os_systems_users_password_override | default('$6$sjH0EeRKZdWYc1y8$F79INhPIfwVUfZ/DRrk.rYEetacoxoqxMVSVRavjggro84vpCqcxUhcNs79C5y4RmdLBb/3sA3vRagfNVjMxD.') }}"
```
manics: I've discovered there's also https://github.com/openmicroscopy/management_tools/blob/master/ansible/inventory/host_vars/outreach.openmicroscopy.org.yml#L1
Presumably the original intention was that the password in this public file could be well known, and we'd override it with the linked private password. If this causes too much confusion we could delete the private override?

manics:
For the benefit of everyone else when @pwalczysko deployed this yesterday it undid the change I made in https://github.com/openmicroscopy/prod-playbooks/pull/74 and broke omero-web due to a duplicate config.

manics:
This change works as expected.

joshmoore:
gh-74 separately merged. Merging the password change now.

omero/training-server/playbook.yml
pwalczysko: I do not hink this is a matter of confusion, rather development of the ecosystem.

Will update the pwd in the MT as well.

PR #87, opened 2018-07-11
Allow www failures

sbesson:
See also https://github.com/openmicroscopy/www.openmicroscopy.org/pull/239

The upstream openmicroscopy.jekyll-build is currently broken as per upstream changes and we are looking into alternate deployment mechanisms. In the meantime, this is allowing the www playbook to fail in Travis so that other valid PRs can be integrated.

Once the deployment strategy is sorted either by using the new artifact strategy or by fixing jekyll-build, Travis should obviously be reenabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ install:
   - pip install ome-ansible-molecule-dependencies
   - pip install jmespath
 
+matrix:
+  allow_failures:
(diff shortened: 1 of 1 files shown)


sbesson:
Closing as per https://github.com/openmicroscopy/prod-playbooks/pull/88

PR #90, opened 2018-07-16
Update to dev django-prometheus role

manics:
This is a required bug-fix for OMERO.web servers using https://github.com/openmicroscopy/ansible-role-omero-web-django-prometheus, and fixes a race condition where an upgraded omero-web may be run before the temporary prometheus stats dir exists.

This fix should be applied before upgrading OMERO.web (run playbook `omero/omero-monitoring-agents.yml`), otherwise OMERO.web may fail to start after the upgrade.

- [ ] --depends-on https://github.com/openmicroscopy/ansible-role-omero-web-django-prometheus/pull/1

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -42,8 +42,9 @@
   version: 0.1.0
 
 - name: openmicroscopy.omero-web-django-prometheus
-  src: https://github.com/openmicroscopy/ansible-role-omero-web-django-prometheus/archive/0.1.0.tar.gz
-  version: 0.1.0
+  # https://github.com/openmicroscopy/ansible-role-omero-web-django-prometheus/pull/1
+  src: https://github.com/manics/ansible-role-omero-web-django-prometheus/archive/984c52bea6ff81713bb30630e9a466869651a827.tar.gz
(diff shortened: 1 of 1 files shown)


manics:
pub-omero was upgraded with this: https://trello.com/c/mzfZdSNi/19-rc2-deployment-testing#comment-5b4ca7eff35139a7ca6b1fb6

kennethgillen:
@manics - what testing do you need/want for this?

manics:
That you can upgrade OMERO.web successfully when the old version of this role is present, and that django metrics still work.

kennethgillen:
Testing also during ns-webclients upgrade:

```
openmicroscopy.omero-web-django-prometheus (984c52bea6ff81713bb30630e9a466869651a827) was installed
```

kennethgillen:
I've not check Prometheus, but if you're getting data from ns-web, ns-web-pub and ome-dundeeomero, consider this good to merge. (I've deployed this to all).

kennethgillen:
👍 upgrades worked after applying, as mentioned previously, and Prometheus showing all green.

PR #91, opened 2018-07-16
Disable openmicroscopy.omero-user (root password unknown)

manics:
This prevents the playbook failing on pub-omero.


kennethgillen:
How about we find out the correct root user, and update `pass` accordingly?

manics:
Yes, this can be closed once it's updated.

joshmoore:
I'd think it's safe for either of you to just reset the password and update credentials.

kennethgillen:
Password updated, and it's now matching what's in credentials.

PR #92, opened 2018-07-16
Pwd required false and email config for web

pwalczysko:
cc @manics @kennethgillen 

this temporary change is necessary for testing of RC2 see [lines 3 and 5](https://docs.google.com/spreadsheets/d/1BMbK1LNxP47ONT7Hj4bCGpY0Soo008jzLVP0_4ZlUTw/edit#gid=774711703).

The playbook should be re-run only against pub-omero which is used for the RC2 testing.

The change for the demo server will be done in another PR later on.

### DIFF
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -356,7 +356,6 @@
     filesystem: "xfs"
 
     omero_server_config_set:
-      omero.security.password_required: false
       omero.db.poolsize: 60
       omero.jvmcfg.percent.blitz: 50
       omero.jvmcfg.percent.indexer: 20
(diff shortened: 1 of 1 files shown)


manics:
This is deployed on pub-omero

kennethgillen:
@pwalczysko if you want to keep pub-omero set up like this, we could set up a separate set of configuration for pub-omero v.s. ome-demoserver like we do for the two Nightshade webclients, `ns-web` and `ns-web-pub`. 

See https://github.com/openmicroscopy/prod-playbooks/blob/master/nightshade-webclients.yml#L207

Otherwise, we can't merge this, given we don't actually want to apply https://github.com/openmicroscopy/prod-playbooks/pull/92/files#diff-896819b5bdf43f197f838c7cc66babf4R359 to `ome-demoserver` during next week's 5.4.7 deployment.

kennethgillen:
Thanks @pwalczysko.

PR #96, opened 2018-08-10
New outreach changes

pwalczysko:
This PR should prep the outreach for the Basel MIAP.

First steps are

  - upgrade to 5.4.7
  - add config steps to iviewer so that Dataset and Well can be opened with ``Open with`` in iviewer

Note that the latter change changes the config for several servers, as there is a shared config file. But I propose this is a harmless, even beneficial change.

cc @jburel @joshmoore @manics

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -282,8 +282,8 @@
     omero_server_system_managedrepo_group: managed_repo_group
     omero_server_datadir_managedrepo_mode: u=rwX,g=srwX,o=rX,+t
     omero_server_datadir_chown: True
-    omero_server_release: "{{ omero_server_release_override | default('5.4.6') }}"
-    omero_web_release: "{{ omero_web_release_override | default('5.4.6') }}"
+    omero_server_release: "{{ omero_server_release_override | default('5.4.7') }}"
+    omero_web_release: "{{ omero_web_release_override | default('5.4.7') }}"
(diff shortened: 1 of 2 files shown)


joshmoore:
What other servers will be impacted, @pwalczysko ?

cc: @kennethgillen

joshmoore:
(Note: restarted some travis jobs due to galaxy timeouts)

pwalczysko:
I think the correct answer is ``nightshade and demo`` if I read correctly the result of the search https://github.com/openmicroscopy/prod-playbooks/search?q=omero-web-config-for-webapps.j2&unscoped_q=omero-web-config-for-webapps.j2 @joshmoore

joshmoore:
Makes sense, @pwalczysko. With that as a general heads up for @kennethgillen et al. during the next upgrade cycle, no objections.

@jburel @manics?

jburel:
Makes sense to do it next upgrade (before end of the month cf. figure 4.0.1)

manics:
The full outreach playbook needs to be run soon anyway to refresh the let's encrypt certificate

joshmoore:
No objections in #sysadmin. Merging.

kennethgillen:
Does iviewer work with `dataset` and `well` out of the box, or does it need a corresponding bump in the version?

jburel:
no need to bump the  version

kennethgillen:
Thanks @jburel

PR #101, opened 2018-08-24
Add minimal ansible.cfg so repo can be run standalone

manics:
In theory this repository can be used standalone without our private config. In practice it is assumed the roles are all in a common directory. This PR therefore adds a minimal ansible.cfg file to control the roles directory (`vendor/`)

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,0 +1,16 @@
+# Ansible configuration file
+# If ansible is executed from this directory it will automatically load
+# this file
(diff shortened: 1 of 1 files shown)


joshmoore:
File works fine with `ansible-galaxy install -r`. `ansible-playbook --list-hosts site.yml` of course shows no hits. I'm fine with adding this; the only potential downside I could think of is if someone is confused that `--list-hosts` is empty, not realizing that s/he is in the wrong directory. Seems minor enough though.

joshmoore:
Taking no objections as a general sign-off & merging.

PR #102, opened 2018-08-24
Outreach prep continued

pwalczysko:
Further prep of the outreach server

   - restructure server script directories (consulted with @will-moore and @jburel, see screenshots below)
   - install matplotlib for the FRAP script
   - do not try to chown omero dir as we have now nfs read-only mounted bits in it (in place imports from mounted idr)
   - bump omero web role, consulted with @manics 

screen shot 2018-08-24 at 16 10 29

Note: here, the content of ``workshop_scripts`` is shown
screen shot 2018-08-30 at 21 27 05

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -150,21 +150,21 @@
         group: "omero-server"
         force: yes
 
-    - name: Download the batch_roi_export_to_table.py script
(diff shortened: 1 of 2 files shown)


jburel:
Merging version used for Basel
Tagging the overarching repo

PR #107, opened 2018-09-20
iptables: Restricting to correct vlan, and adding GPFS port.

kennethgillen:
Further tightening of the firewall rules, and addition of a required `GPFS` port.

### DIFF
--- a/omero/omero-firewall.yml
+++ b/omero/omero-firewall.yml
@@ -49,7 +49,8 @@
   # All other ports that allow incoming connections:
   # - web
   # - omero
-  # - checkmk
+  # - GPFS
+  # - Check_MK
   - name: Iptables OME ports
     become: yes
     iptables_raw_25:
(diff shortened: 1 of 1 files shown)


kennethgillen:
https://github.com/openmicroscopy/prod-playbooks/pull/107/files#diff-c4fd213fe44fcc72f487cf67bf9a6aa5R62 has been tested on ome-demoserver by running a rule-replace

```
[root@ome-demoserver ~]# iptables -L INPUT 7
ACCEPT     tcp  --  anywhere             anywhere             tcp dpt:6556 /* ansible[ome_ports] */
[root@ome-demoserver ~]# iptables -R INPUT 7 -p tcp -m tcp --dport 6556 -s 10.2.1.0/24 -j ACCEPT
[root@ome-demoserver ~]# iptables -L INPUT 7
ACCEPT     tcp  --  vl34.net.lifesci.dundee.ac.uk/24  anywhere             tcp dpt:6556
```
Then re-running the Check_MK check which uses this TCP port.

Then made the change permanent by hand-editing `/etc/sysconfig/iptables`.

manics:
As a quick fix this looks OK. Longer term we should probably separate out the GPFS config because it only applies to one server whereas these rules are applied to all servers.

Why did you need to do a manual change? Running this modified playbook against demo should've have handled everything.

kennethgillen:
It was run manually as I had to do it in a hurry, and I didn't have time to work out how the Ansible was implemented.

PR #115, opened 2018-10-12
Bump OMERO.web systemd number of files

kennethgillen:
This has been an outstanding TODO for a while.

It was applied to OMERO.server, but OMERO.web was missed.


manics:
Looks OK to me

kennethgillen:
@joshmoore - is ^ sufficient review for a merge?

joshmoore:
Works for me as well. I'd leave it to you to check the actual before/after value when you deploy next.

PR #116, opened 2018-10-16
SSL update for legacy hosts.

kennethgillen:
A playbook to deploy SSL certs to legacy hosts.

Supporting MT configuration, including an inventory update at: https://github.com/openmicroscopy/management_tools/pull/864

### DIFF
--- a/legacy/ssl-deploy.yml
+++ b/legacy/ssl-deploy.yml
@@ -1,0 +1,24 @@
+---
+# k.h.gillen@dundee.ac.uk Mon 15 Oct 2018 11:11:42 BST
+
+# Playbook which deploys SSL certificates to legacy
+# (not entirely Ansible managed) hosts
+
+# openmicroscopy.org wildcard
+- hosts: legacy-ssl-org-wildcard
+  roles:
+  - role: openmicroscopy.ssl-certificate
+    # Vars here since vars will change for each certificate we're deploying.
+    ssl_certificate_public_content: "{{ ssl_org_wildcard_certificate_public_content }}"
+    ssl_certificate_intermediate_content: "{{ ssl_org_wildcard_certificate_intermediate_content }}"
+    ssl_certificate_key_content: "{{ ssl_org_wildcard_certificate_key_content }}"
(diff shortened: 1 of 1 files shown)


legacy/ssl-deploy.yml
```diff
@@ -0,0 +1,24 @@
...
+
+# Playbook which deploys SSL certificates to legacy
+# (not entirely Ansible managed) hosts
+
+# openmicroscopy.org wildcard
+- hosts: legacy-ssl-org-wildcard
+  roles:
+  - role: openmicroscopy.ssl-certificate
+    # Vars here since vars will change for each certificate we're deploying.
+    ssl_certificate_public_content: "{{ ssl_org_wildcard_certificate_public_content }}"
```
manics: Optional: You could avoid having multiple variables by omitting this variable from the playbook and defining it in host/group vars only, e.g.:
- https://github.com/openmicroscopy/management_tools/blob/594a39d85dfd25869382a9da0b409a8c901a0adf/idr/inventory/group_vars/proxy-hosts.yml#L5

kennethgillen: I envisage a second instance of this same role in this same playbook, for the `.org.uk` wildcard.

So in this case if `ssl_certificate_public_content` was defined in two hostgroups, and the same host was in both groups, I don't know if Ansible would cope with that. i.e. how would it know which to use for each invocation? Or is there obvious way to work around/apply that pattern that I've missed?

kennethgillen:
Deployed to `ci-hosts`: 

```
PLAY RECAP **************************************************************************************************
cowfish.openmicroscopy.org : ok=10   changed=4    unreachable=0    failed=0
octopus.openmicroscopy.org : ok=10   changed=4    unreachable=0    failed=0
ray.openmicroscopy.org     : ok=10   changed=4    unreachable=0    failed=0
seabass.openmicroscopy.org : ok=10   changed=4    unreachable=0    failed=0
trout.openmicroscopy.org   : ok=0    changed=0    unreachable=1    failed=0
```

kennethgillen:
This is ready to merge.

PR #117, opened 2018-10-19
Website update

jburel:
Update website to version 2018.10.19
changes made to update www.o.org and ome-www.dev.o.org


sbesson:
Deployed live already.

PR #121, opened 2018-11-12
Adding playbook for a standalone PostgreSQL server.

kennethgillen:
This is a replacement for https://github.com/openmicroscopy/infrastructure/pull/302

Inventory and host vars already merged at https://github.com/openmicroscopy/management_tools/pull/540/files

### DIFF
--- a/pg-int-testing.yml
+++ b/pg-int-testing.yml
@@ -1,0 +1,113 @@
+# Install PostgreSQL and prepare the OME (UoD/SLS) prerequisites
+
+- hosts: pg-int-testing.openmicroscopy.org
+  pre_tasks:
+    - name: Install open-vm-tools if system is a VMware vm
+      become: yes
+      yum:
+        name: open-vm-tools
+        state: latest
+      when: >
+           ((ansible_virtualization_type is defined)
+           and (ansible_virtualization_type == "VMware"))
+
+    # Perhaps alter the role at https://github.com/openmicroscopy/ansible-role-lvm-partition/
+    # to make some of the variables non-required.
+    - name: Resize root FS without altering mount options
+      tags: lvm
+      become: yes
+      lvol:
+        lv: root
+        vg: VolGroup00
+        size: "{{ provision_root_lvsize }}"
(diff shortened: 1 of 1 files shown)


kennethgillen:
>Did openmicroscopy/infrastructure#302 (comment) get addressed?
No. That code was to make the install more robust in a situation I didn't find the source of - that of why some agent installs didn't "ship with" the plugin. If we want to remove those conditionals, in my view it only makes the deploy more brittle, as the source of the issues is as yet undiscovered. Also, given that code was already deployed to many other production systems, I just went with "what works". As you say, the whole check_mk plugin/agent install should be refactored out into a role.

>Should the name be widened to include other PG boxes?
Potentially yes. I am spinning up a temporary one here using the same playbook, but didn't want to make it generic to support that, given it's one I'll be blowing away after use.

> (additionally why is pg-int-testing in prod-hosts?)
It's what I think of a production for the team, but not production for external services.

kennethgillen:
I notice the postgresql LVM LV isn't created by this playbook, so we might want to hold ON_HOLD till I add that.

kennethgillen:
undoing `--exclude` label - since adding the postgresql LVM config for the VolumeGroup used here is beyond the scope of the original playbook. That piece was done manually.

Given each new DB server may have a differing disk layout (SSD/not, new VolumeGroup, etc) probably best leaving this as a manual bootstrapping step.

joshmoore:
> No. That code was to make the install more robust in a situation I didn't find the source of ...

Ok. Anything that needs capturing then?

> Potentially yes. I am spinning up a temporary one here using the same playbook, but didn't want to make it generic to support that, given it's one I'll be blowing away after use.
> ...
> It's what I think of a production for the team, but not production for external services

Guess my biggest concern is that this repo is "prod-playbooks", so the naming should mean something to someone coming in, e.g. `databases.yml` or `pg-servers.yml`?

sbesson:
Closing as per no activity. The need for a standalone PSQL playbook might come back and we can reuse this PR as a base.

PR #127, opened 2018-12-02
aliases: handle uppercase product names

joshmoore:
Since our logos are all uppercase, placing "openmicroscopy.org" in
front of them might encourage someone to go, for example,
openmicroscopy.org/OMERO which currently is 404.

This has been run on ome-www-dev with:

```
ansible-playbook -i ../ansible/inventory/staging-hosts  playbooks/www/www-deploy.yml --ask-become
```

Test with:
 * https://ome-www-dev.openmicroscopy.org/OMERO/
 * https://ome-www-dev.openmicroscopy.org/BIO-FORMATS/downloads
 * etc.

### DIFF
--- a/www/www-deploy.yml
+++ b/www/www-deploy.yml
@@ -37,6 +37,12 @@
 
     nginx_proxy_redirect_map_locations:
     # TODO: change to 301 when we're happy
+    - location: "~ ^/(BIO-FORMATS)($|/)"
+      code: 302
+    - location: "~ ^/(OME-FILES)($|/)"
+      code: 302
+    - location: "~ ^/(OMERO)($|/)"
+      code: 302
     - location: "~ ^/(site)($|/)"
       code: 302
     - location: "~ ^/(omero-blog)($|/)"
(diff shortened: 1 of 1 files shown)


www/www-deploy.yml
```diff
@@ -202,6 +208,20 @@
...
+      dest: /bio-formats
+    - match: "~/BIO-FORMATS/(?<link>.*)$"
+      dest: /bio-formats/$link
+    - match: "~/OME-FILES$"
+      dest: /ome-files
+    - match: "~/OME-FILES/(?<link>.*)$"
+      dest: /ome-files/$link
+    - match: "~/OMERO$"
+      dest: /omero
+    - match: "~/OMERO/(?<link>.*)$"
```
sbesson: My primary questions was whether we need the full range of regexp handling in which case this PR is definitely the way to go or only handle capitalized versions for some defined URLS e.g. `/OMERO/` and maybe `/OMERO/downloads`. In the latter case, I would be including to use `jekyll-redirect-from` in the usptream http://github.com/openmicroscopy/www.openmicroscopy.org code for single-page internal redirect.

sbesson:
I have been testing this alongside #167. I don't see a reason not to include it, we can always review assumptions later

PR #128, opened 2018-12-14
Add config to enable parallel imports reliably

pwalczysko:
cc @sbesson @mtbc 

This addresses the problem of two concurrent imports under the same user  failing if they try to create one of the following folders under the ManagedRepository at the same time:

See https://github.com/IDR/idr0021-lawo-pericentriolarmaterial/pull/3#issuecomment-447126321 - this PR is going for the second suggested solution (by @sbesson ) because it is more robust solution I think.

I suppose there will be no detrimental effects coming from the fact that some data on the server were imported with the old config setup, whilst some (new, incoming) data will be imported using the path config in this PR ?

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -329,6 +329,7 @@
       omero.jvmcfg.percent.blitz: 50
       omero.jvmcfg.percent.indexer: 20
       omero.jvmcfg.percent.pixeldata: 20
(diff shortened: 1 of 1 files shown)


mtbc:
OMERO should be fine with its data having been imported with a variety of different kinds of repository path.

sbesson:
Merging as per the successful functional testing mentioned in https://github.com/IDR/idr0021-lawo-pericentriolarmaterial/pull/3#issuecomment-447391159

PR #134, opened 2019-01-24
Bump version of OME website

sbesson:
Version 2019.01.24 includes the announcement for the OMERO 5.4.10 release


sbesson:
Note this has been deployed on staging and prod - see https://www.openmicroscopy.org/2019/01/24/omero-5-4-10.html

PR #137, opened 2019-02-04
Add more formats with public subfolders

sbesson:
New formats are a combination of new formats supported in BF 6 (KLB) and existing formats which were not mapped at this level yet.

Names of the format folders are up for discussion /cc @openmicroscopy/formats


joshmoore:
In general, :+1: for getting more public. Obviously the naming angle is interesting. i.e. systematic or "generally known as". Wonder if that should also somehow relate to the generated docs for the given reader.

sbesson:
Merging and deploying. We can certainly review the naming of the individual formats. For new formats, we try to match the naming used in the https://docs.openmicroscopy.org/bio-formats/5.9.2/supported-formats.html table but many of these are decide quite arbitrarily. If we need to update, creating alias is always a possibility.

PR #141, opened 2019-02-22
Nginx proxy role bump (www)

manics:
The role reverts the breaking log format change, so this is now defined as a parameter.
- See https://github.com/openmicroscopy/ansible-role-nginx-proxy/pull/21

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -25,9 +25,10 @@
 - src: openmicroscopy.nginx
   version: 1.0.0
 
-- src: openmicroscopy.nginx-proxy
-  version: 1.8.0
-
+# https://github.com/openmicroscopy/ansible-role-nginx-proxy/pull/21
+- name: openmicroscopy.nginx-proxy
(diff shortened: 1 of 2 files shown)


sbesson:
Thanks for opening the PR. I am not sure the custom NGINX format is required for the `www` server which has only 1 host. No objection to dropping 24d0745

manics:
`ome-www` already has the new format from `1.8.0`, I can revert it but it means there'll have been an intervening period where the logs would've been in the other format. If we're not doing any log analysis on that box that won't matter.

sbesson:
Unfortunately, we do already have a mixture of logs into 2 formats on this server. I do not know the `ome-www` logs are analyzed at the moment otherwise we would probably have flagged the log format change. If we do so, we will need to handle both variations.
Unless @joshmoore remembers specific reason not to do so, my vote still goes towards reverting the log format to using the default format.

manics:
No one objected so I've reverted the last commit

sbesson:
https://galaxy.ansible.com/openmicroscopy/nginx-proxy `1.9.0` is now tagged

sbesson:
Thank you. Merging and proposing to schedule the redeployment of www tomorrow with the new format.

PR #145, opened 2019-03-18
Demo cyberessentials

manics:
Update Nginx to 1.14.2
Update SSL settings including HSTS
Update OMERO.web cookie and related settings
Update the playbook to use the postgresql ansible role 3 (changes default timezone from `GB` to `UTC`

Since this includes a minor refactoring you must delete `/etc/nginx/conf.d/omero-web.conf` to force it to be regenerated. See https://github.com/openmicroscopy/prod-playbooks/commit/02b7d94349d80dbbcbe16356d0d382aad2413958

SSL protocols and ciphers can be verified using https://nmap.org/nsedoc/scripts/ssl-enum-ciphers.html
`nmap -sV --script ssl-enum-ciphers.nse -p 443 HOSTNAME`

### DIFF
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -360,3 +360,6 @@
     ssl_certificate_key_path: /etc/nginx/ssl/server.key
     # Server path to SSL combined certificate and key, set to empty to disable
     ssl_certificate_combined_path: ''
+
+
+- import_playbook: omero/omero-web-patch-settings.yml
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -110,6 +110,10 @@
         password: "{{ vault.omero_server_dbpassword }}"
         databases: ["{{ vault.omero_server_dbname }}"]
 
+    - role: openmicroscopy.nginx
+      # Defaults overridden in private configuration
+      # nginx_version:
+
     - role: openmicroscopy.omero-server
       omero_server_release: "{{ omero_server_release_desired | default('5.4.9') }}"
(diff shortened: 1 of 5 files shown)


manics:
This is deployed to pub-omero.

manics:
This is deployed to pub-omero. OpenSSL was upgraded yum upgrade openssl but doesn't seem to have any noticeable effects.

joshmoore:
Generally :+1: since this is restricted to `demo`-like systems, but:
 * Mentioned var naming elsewhere
 * :+1: for the temporary solution assuming it gets refactored upstream
 * but "Since this includes a minor refactoring you must delete /etc/nginx/conf.d/omero-web.conf to force it to be regenerated." concerns me a bit. Will this persist? Will it impact users?

manics:
> "Since this includes a minor refactoring you must delete /etc/nginx/conf.d/omero-web.conf to force it to be regenerated." concerns me a bit. Will this persist? Will it impact users?

This is a consequence of using multiple non-deterministic tasks to update the omero-web.conf file generated by `omero web config`- it's very difficult to undo a change, so forcing the file to be recreated from scratch is easier and more reliable. The easiest solution is to avoid running the `omero web config` command and instead bundle an Ansible configuration template in the omero-web Ansible role (or alternatively bundle the template in OMERO.py)

Related discussion: https://trello.com/c/UzWEvrPE/119-improve-omeroweb-default-security

manics:
Deployed to demo

PR #150, opened 2019-03-21
Nightshade and nswebclients cyberessentials

manics:
Similar to the changes made on demo: https://github.com/openmicroscopy/prod-playbooks/pull/145
- Update Nginx to 1.14.2
- Update SSL settings including HSTS
- Update OMERO.web cookie and related settings
- In addition adds the `/pub/schleicher-et-al-2017` → `/webclient/?show=project-27936` redirect on both servers

Since this includes a minor refactoring you must delete `/etc/nginx/conf.d/omero-web.conf` on `ns-webclients` to force it to be regenerated:
- [ ] delete `ns-web:/etc/nginx/conf.d/omero-web.conf`
- [ ] delete `ns-web-pub:/etc/nginx/conf.d/omero-web.conf`

Note there are three Nginx servers and two OMERO.web servers:
- ns-web (internal read-write OMERO.web + Nginx)
- ns-web-pub (omero.lifesci, public read-only OMERO.web + Nginx)
- nightshade (Nginx only, proxied to ns-web)

### DIFF
--- a/nightshade-webclients.yml
+++ b/nightshade-webclients.yml
@@ -27,6 +27,8 @@
       lvm_lvsize: "{{ provision_rootsize }}"
       lvm_lvfilesystem: "{{ provision_root_filesystem }}"
 
+    - role: openmicroscopy.nginx
+
     # OMERO.web configuration in host_vars in different repository
     - role: openmicroscopy.omero-web
       omero_web_release: 5.4.9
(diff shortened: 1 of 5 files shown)


manics:
In check mode there is one unexpected change to nightshade:
```
TASK [openmicroscopy.omero-server : omero server | systemd service] ************
--- before: /etc/systemd/system/omero-server.service
+++ after: /Users/spli/.ansible/tmp/ansible-local-29892JTJ8Vb/tmppLHTHM/systemd-system-omero-server-service.j2
@@ -8,6 +8,8 @@
 After=postgresql-9.6.service
 Requires=network.service
 After=network.service
+After=gpfs.service
+Requires=gpfs.service
 
 [Service]
 User=omero-x

changed: [ome-dundeeomero.openmicroscopy.org]
```
Looks like https://github.com/openmicroscopy/prod-playbooks/pull/59/ was never applied.

PR #151, opened 2019-03-21
demoservers: use default ssl_certificate paths from role

manics:
Use the role default `/etc/ssl/localcerts/` for all certs.


manics:
This is deployed to demo and pub-omero

PR #152, opened 2019-03-22
Nightshade cyberessentials nginx workarounds

manics:
ns-web and ns-web-pub were successfully upgraded with https://github.com/openmicroscopy/prod-playbooks/pull/150
nightshade failed due to a failure to update Nginx dependencies, so this PR reverts commit https://github.com/openmicroscopy/prod-playbooks/commit/e01f143472951a675d97993d7588d2c66f5a9854 and overrides the ssl ciphers.

This has been deployed


templates/nginx-omero.conf.j2
```diff
@@ -4,8 +4,27 @@ server {
 }
 
 server {
+    listen 443 ssl;
     server_name $hostname;
 
+    ssl_certificate {{ ssl_certificate_bundled_path }};
+    ssl_certificate_key {{ ssl_certificate_key_path }};
+    ssl_protocols   TLSv1 TLSv1.1 TLSv1.2;
```
mtbc: While we're at it should we just drop at least `TLSv1`? (Or do we not actually support it?)

manics: The current list matches the latest Nginx default: http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_protocols

manics:
Already deployed

PR #159, opened 2019-04-23
Add redirect for new link to publication on nightshade pub

pwalczysko:
Adding a redirect for the new publication group on nightshade.

cc @sbesson


pwalczysko:
@sbesson I think I might be hitting something like https://stackoverflow.com/questions/34803467/unexpected-exception-name-basestring-is-not-defined-when-invoking-ansible2  here ? Not sure how to solve though at all

PR #160, opened 2019-04-23
Set Python to 2.7 in .travis.yml

sbesson:
See https://github.com/openmicroscopy/prod-playbooks/pull/159#issuecomment-485848708

`ome-ansible-molecule-dependencies` is shipping Ansible 2.4 which requires Python 2.7


pwalczysko:
Thank you @sbesson

PR #166, opened 2019-05-23
Install git on outreach

pwalczysko:
Discussed with @sbesson : Git installation on outreach is of advantage to be able to manage the versions of the training scripts there (maintenance mainly) .

Second goal here is to be able to clone idr repos and thus use the bulk.yml and tsv files from there easily.

@jburel @manics @joshmoore

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -49,6 +49,8 @@
         databases: [omero]
       postgresql_version: "9.6"
 
+    - role: openmicroscopy.versioncontrol-utils
+
     - role: openmicroscopy.nfs-mount
 
     - role: openmicroscopy.omero-server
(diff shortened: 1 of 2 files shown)


pwalczysko:
``mERROR! the role 'openmicroscopy.versioncontrol-utils' was not found in /home/travis/build/ome/prod-playbooks/omero/training-server/roles:/tmp/molecule/prod-playbooks/omero-training-``

hum, what now - ``galaxy`` ? fixing travis ? @sbesson

sbesson:
If it's not defined it probably need to be added to the requirements file - https://github.com/ome/prod-playbooks/blob/master/requirements.yml

pwalczysko:
went for the last version as found on https://github.com/openmicroscopy/ansible-role-versioncontrol-utils/releases

PR #167, opened 2019-05-28
www: redirect /forums to image.sc

joshmoore:
see: https://github.com/ome/www.openmicroscopy.org/pull/295

cc: @manics @sbesson @jburel


sbesson:
f7b266a should fix the redirect. This has now been deployed on staging so that https://ome-www-dev.openmicroscopy.org/forums redirect to the image.sc forum

The following commits start migrating the original manual redirect tests written during the `www` migration as Molecule tests (including the new redirects). This should add extra time in Travis but ensure coverage of the Web server behavior as we modify it. Feedback welcome.

PR #168, opened 2019-06-06
Propose new forums location: data-management

joshmoore:
For discussion: Now that /support explicitly lists tags (https://github.com/ome/www.openmicroscopy.org/pull/298), perhaps the overall link can go to the new category?


sbesson:
Works for me.

sbesson:
Now deployed live

joshmoore:
Wow, thanks! (I like it...)

PR #173, opened 2019-06-17
install zsh

jburel:
Install zsh on outreach

In checkmode
```
TASK [Prerequisites for CLI] *************************************************************************************************************************************************
changed: [outreach.openmicroscopy.org] => (item=[u'zsh'])


```

cc @dominikl

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -32,6 +32,14 @@
         - gcc
         - python-ldap
 
+    - name: Prerequisites for CLI
+      become: yes
+      yum:
+        name: "{{ item }}"
+        state: present
+      with_items:
+        - zsh
+
   roles:
 
     - role: openmicroscopy.postgresql-3


omero/training-server/playbook.yml
```diff
@@ -32,6 +32,14 @@
         - gcc
         - python-ldap
 
+    - name: Prerequisites for CLI
```
sbesson: :+1: for handling via tasks first. If this becomes handy, a question is whether we would like to have this systematically installed as part of https://github.com/ome/ansible-role-basedeps/blob/master/vars/main.yml? /cc @manics

dominikl:
Thanks JM 👍 I think as zsh is compatible to bash and apparently will become the default shell for OSX too, it would be good to have it available on all systems.

omero/training-server/playbook.yml
manics: It's not really a basedep in the sense of something that's required for all OMERO systems. It could potentially go into a new role for packages that we want on all our production systems.

joshmoore: @manics : do you have any other suggestions for such a role? (And a name?)

manics: `ome-ui-tools`, `ome-cli-tools`, `ome-dev-tools`?

sbesson: While migrating roles, I (re-)discovered https://github.com/openmicroscopy/ansible-role-cli-utils, would this be a useful addition there ?

jburel: looking at https://github.com/openmicroscopy/ansible-role-cli-utils/blob/master/tasks/main.yml
this sounds like a good candidate

jburel: see https://github.com/openmicroscopy/ansible-role-cli-utils/pull/3

jburel:
I will open a new PR now that we have the correct role updated.

PR #174, opened 2019-06-17
Nightshade web

jburel:
* Remove usage of ``upgrade_webapps`` parameter
* Usage variables

```
PLAY RECAP *******************************************************************************************************************************************************************
ns-web.openmicroscopy.org  : ok=77   changed=6    unreachable=0    failed=0   
```

### DIFF
--- a/nightshade-webclients.yml
+++ b/nightshade-webclients.yml
@@ -126,7 +126,7 @@
     omero_webtagging_autotag_release: "{{ omero_webtagging_autotag_release_override | default('3.0.2') }}"
     omero_webtagging_tagsearch_release: "{{ omero_webtagging_tagsearch_release_override | default('3.0.3') }}"
     omero_web_release_version: "{{ omero_web_release_version_override | default('5.4.10') }}"
-
+    nginx_version: 1.14.2
     # Disable SSL combined certificate and key
     ssl_certificate_combined_path: ''
 
--- a/nightshade-webclients.yml
+++ b/nightshade-webclients.yml
@@ -125,7 +125,7 @@
     omero_parade_release: "{{ omero_parade_release_override | default('0.1.2') }}"
     omero_webtagging_autotag_release: "{{ omero_webtagging_autotag_release_override | default('3.0.2') }}"
     omero_webtagging_tagsearch_release: "{{ omero_webtagging_tagsearch_release_override | default('3.0.3') }}"
-    omero_web_release_version: "{{ omero_web_release_version_override | default('5.4.10') }}"
+    omero_web_release_version: "{{ omero_web_release_version_override | default('5.5.0') }}"
     nginx_version: 1.14.2
     # Disable SSL combined certificate and key
     ssl_certificate_combined_path: ''
--- a/nightshade-webclients.yml
+++ b/nightshade-webclients.yml
@@ -20,7 +20,7 @@
 
     # OMERO.web configuration in host_vars in different repository
     - role: openmicroscopy.omero-web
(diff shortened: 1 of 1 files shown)


nightshade-webclients.yml
```diff
@@ -145,8 +119,13 @@
 
   vars:
     omero_web_config_set: "{{ omero_web_config_set_for_group | combine(omero_web_config_set_for_host) }}"
-
-    nginx_version: 1.14.2
```
manics: What's the reason for removing `nginx_version`?

nightshade-webclients.yml
```diff
@@ -145,8 +119,13 @@
...
+    omero_figure_release: "{{ omero_figure_release_override | default('4.0.2') }}"
+    omero_fpbioimage_release: "{{ omero_fpbioimage_release_override | default('0.3.0') }}"
+    omero_iviewer_release: "{{ omero_iviewer_release_override | default('0.7.0') }}"
+    omero_parade_release: "{{ omero_parade_release_override | default('0.1.2') }}"
+    omero_webtagging_autotag_release: "{{ omero_webtagging_autotag_release_override | default('3.0.2') }}"
+    omero_webtagging_tagsearch_release: "{{ omero_webtagging_tagsearch_release_override | default('3.0.3') }}"
+    omero_web_release_version: "{{ omero_web_release_version_override | default('5.4.10') }}"
```
manics: Are you planning to upgrade Nightshade OMERO.web to 5.4.10, and the web-apps too?
Current versions on ns-web:
OMERO.py: 5.4.9
```
$ /opt/omero/web/venv/bin/pip list |grep omero
omero-figure (4.0.0)
omero-fpbioimage (0.3.0)
omero-iviewer (0.6.0)
omero-marshal (0.5.3)
omero-parade (0.1.2)
omero-webtagging-autotag (3.0.2)
omero-webtagging-tagsearch (3.0.3)
```

jburel: Everything needs to be upgraded. apps and web
If people are ok with change, I will go directly to 5.5.0 for web

nightshade-webclients.yml
jburel: we did not specify for example in outreach

manics: This was introduced during the cyberessentials work as the Nginx verison was out of date: https://github.com/ome/prod-playbooks/pull/150/ . I think it should be brought back so we have full control over the version.

nightshade-webclients.yml
manics: I think it's important that when a PR is merged it represents the current state of our systems, so depending on your preference either change the versions to match what's currently deployed, or include the 5.5.0 upgrade in this PR.

nightshade-webclients.yml
jburel: understood

nightshade-webclients.yml
jburel: I have only run in ``checkmode`` so nothing has been upgraded. 
In my view, It is preferable to have any changes reviewed before running without ``checkmode`` if possible

joshmoore: Agreed. With nightshade it's more difficult, but I understood the general agreement to be:

 * Open a PR with checkmode against staging & prod server output
 * If general agreement, run against staging unmerged
 * If everything looks ok, merge and then rollout to prod

sbesson:
@manics are you happy with the last commits. Should we get this merged in preparation of the OMERO 5.5.1 upgrade changes across prod servers?

PR #175, opened 2019-06-17
nightshade playboook cleanup

jburel:
* remove duplicate mencoder installation
* install ice from wheel,
* do not install ice-devel 

in check mode
```

PLAY RECAP *******************************************************************************************************************************************************************
ome-dundeeomero.openmicroscopy.org : ok=75   changed=5    unreachable=0    failed=0   
```

### DIFF
--- a/ome-dundeeomero.yml
+++ b/ome-dundeeomero.yml
@@ -249,8 +249,6 @@
     #omero_server_db_dumpdir_name: nightly-pg_dump_omero.dir
 
     ice_version: "3.6"
-    ice_install_devel: False
-    ice_python_wheel: "https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.1.0/zeroc_ice-3.6.4-cp27-cp27mu-linux_x86_64.whl"
     postgresql_version: "9.6"
     filesystem: "xfs"
     omero_figure_release: "{{ omero_figure_release_override | default('v4.0.2') }}"
--- a/ome-dundeeomero.yml
+++ b/ome-dundeeomero.yml
@@ -48,12 +48,6 @@
         name:  http://li.nux.ro/download/nux/dextop/el7/x86_64/nux-dextop-release-0-5.el7.nux.noarch.rpm
         state: present
 
-    - name: Install Make Movie script Prerequisite | MEncoder - Package
-      become: yes
-      yum:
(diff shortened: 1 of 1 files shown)


jburel:
demo is ``ome-demoserver.yml``

sbesson:
Changes here are inline with https://github.com/ome/prod-playbooks/pull/171 and https://github.com/ome/prod-playbooks/pull/175. The only difference is the ownership of the `Figure_To_Pdf` script which is `root` here but `omero-server` on `ome-demoserver.yml`. Is there a reason to keep the difference or should we unify?

ome-dundeeomero.yml
```diff
@@ -254,9 +248,12 @@
     #omero_server_db_dumpdir_parent: /tmp
     #omero_server_db_dumpdir_name: nightly-pg_dump_omero.dir
 
+    ice_version: "3.6"
+    ice_install_devel: False
+    ice_python_wheel: "https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.1.0/zeroc_ice-3.6.4-cp27-cp27mu-linux_x86_64.whl"
```
joshmoore: Outwith this PR, it seems unfortunate to need to specify this URL everywhere.

manics: FYI I don't think these changes are strictly necessary, I think the problem may have been due to a change in the zeroc yum repo. Running this in check mode yesterday showed a change to the yum repo, and since this was in check mode the change wouldn't have taken effect, which means the package wouldn't be found. In the absence of a molecule test it's difficult to verify this.

jburel:
@sbesson It was a bit unclear why it was ``root`` in first instance so i did not want to modify for this round. I could change and roll back if need be

ome-dundeeomero.yml
jburel: We should not need ice-devel on nightshade.

manics: We shouldn't, but if it was previously in the playbook then without a clean install (e.g. a molecule test) we can't be sure there isn't anything secretly depending on it. `ice_install_devel: True` is also the default for the role so represents the current state of the server

jburel:
* Playbook run successfully against nightshade
* @mtbc logged in via CLI

PR #172, opened 2019-06-13
demo webapps

jburel:
remove usage of webapp parameter

I ran it changes in check mode

PR #178, opened 2019-06-19
pub-omero/demo: install omero-signup

manics:
Installs and configures https://gitlab.com/openmicroscopy/incubator/omero-signup/tree/v0.0.4
This has been deployed on pub-omero
Depending on when https://github.com/ome/prod-playbooks/pull/172 is deployed I'll either rebase after it's merged, or open this PR on top of it.

If you're interested in what other configuration options are possible see https://gitlab.com/openmicroscopy/incubator/omero-signup/blob/v0.0.4/omero_signup/signup_settings.py#L13-42

Testing: sign-up for a new account at https://pub-omero.openmicroscopy.org/signup/


joshmoore:
Can you add some info on what you think the state of 0.0.4 is an what, for example, it'd take to get to 0.1.0?

manics:
Probably testing by @pwalczysko since he'll probably find a way to break it, and code review, especially concerning the need for the app to have an admin account to create users and groups.

pwalczysko:
Screen Shot 2019-06-19 at 16 56 11

Studying the layout: The encouragement to ``If you already have account, login here`` is placed unfortunately right next to "Create account" button - the two do not belong together, but the layout strongly suggests they do.

snoopycrimecop:
Conflicting PR. Removed from build [TRAINING-merge#305](https://ci.openmicroscopy.org/job/TRAINING-merge/305/). See the [console output](https://ci.openmicroscopy.org/job/TRAINING-merge/305/consoleText) for more details.
Possible conflicts:
  - PR #172 jburel 'demo webapps'
    - ome-demoserver.yml

--conflicts

joshmoore:
In addition to @pwalczysko's UI comment, 

## Basic
 * The incubator project is a mirror of your personal repo: doesn't look like I can create an issue there. (so doing code review here)
 * :+1: for bumpversion (Do you see this as being the basis for other plugins?)
 * "omero.web.signup.host" Could this plugin into the server list somehow? (Or are we getting away from that?)
 * "configuring email on a test server:" you could use the fake SMTP in omero-server (in a docker-compose, I suppose. Shall that be the default for the omero-server in test-infra?)

## Safety
 * email: if it's okay to show the password in the acknowledgement (meaning the email hasn't been validated), then perhaps sending the email could be optional?
 * Is there any rate limiting or similar? (Are you assuming this will be safe since the URL won't be accessible?)

## Minor
 * Want to update the changelog?

## Usability
 * You could probably auto-login the user, no?

## Future / long-term maintainability
 * Where did ome.signup.css ? Are there any issues of keeping this in sync?
 * index.html: the hard-coding of `2007` in the footer seems odd.
 * This makes me wonder if `forgottenpassword` should be a plugin as well.

manics:
> The incubator project is a mirror of your personal repo

I've turned on issues and merge requests on https://gitlab.com/openmicroscopy/incubator/omero-signup/

> 👍 for bumpversion (Do you see this as being the basis for other plugins?)

It's a lot nicer than manually editing versions in multiple files, and tagging.

> "omero.web.signup.host" Could this plugin into the server list somehow? (Or are we getting away from that?)

Could do but how should multiple servers be handled? Add a dropdown to select? Have multiple sets of admin credentials for each? Does that also need to apply to all config options such as the group-name template?

> email: if it's okay to show the password in the acknowledgement (meaning the email hasn't been validated), then perhaps sending the email could be optional?

It's already optional, `omero.web.signup.email.enabled`

> Is there any rate limiting or similar? (Are you assuming this will be safe since the URL won't be accessible?)

No, I can't see how to do that without support for storing state somewhere in OMERO

An alternative to allowing free-for-all signups is to use one or more external OAuth providers: https://gitlab.com/openmicroscopy/incubator/omero-oauth
If you want me to install that too create some credentials following https://gitlab.com/openmicroscopy/incubator/omero-oauth/blob/master/multi-example.yaml and I'll deploy it

joshmoore:
> Could do but how should multiple servers be handled? 

Might be worth a quick discussion. Should plugins _ever_ care about the server concept?

> It's already optional, omero.web.signup.email.enabled

Sorry, optional for the user.

> No, I can't see how to do that without support for storing state somewhere in OMERO

Ok. This is probably my biggest concern about activating it as it stands.

> ...to use one or more external OAuth providers: 

The OAuth step would just to be verify that the user is real? You'd still need to generate a password, right?

manics:
> The OAuth step would just to be verify that the user is real? You'd still need to generate a password, right?

A password is optional ... https://gitlab.com/openmicroscopy/incubator/omero-oauth/blob/v0.0.2/omero_oauth/views.py#L200-212
though this pathway definitely needs code review 😀

joshmoore:
> A password is optional 

Really?! And for import?

manics:
A new view can provide a session token: https://gitlab.com/openmicroscopy/incubator/omero-oauth/blob/v0.0.2/omero_oauth/views.py#L232-258

snoopycrimecop:
Conflicting PR. Removed from build [TRAINING-merge#306](https://ci.openmicroscopy.org/job/TRAINING-merge/306/). See the [console output](https://ci.openmicroscopy.org/job/TRAINING-merge/306/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - ome-demoserver.yml

--conflicts

joshmoore:
> A new view can provide a session token

I don't imagine we want to be hand-holding users through that process...

manics:
Which route would you prefer to proceed with? omero-signup and see if anyone else has ideas for rate-limiting? Or omero-oauth and generate a password automatically?

manics:
Rebased and added [omero-oauth](https://gitlab.com/openmicroscopy/incubator/omero-oauth/tree/v0.0.2): Login here: https://pub-omero.openmicroscopy.org/oauth/

Note there may be some incompatible Python package versions on pub-omero, not sure whether this has accumulated over several updates.
Also there are intermittent `SSLError: ('failed to allocate SSL context',)` errors when signing in to OMERO.web with OAuth. I can't reproduce this locally. Retrying usually works.

joshmoore:
Similar feedback on oauth:
 * similar multiple server issue & css issue
 * not a huge fan of the external yaml file due to potential upgrade loss, but :+1: for supporting both
 * :+1: for providers.py. Are there enough of these that eventually it needs to be an extension point?
 * changelog & headers. (Does any of the code have external copyrights?)
 * might be good to have tests against test-infra
 * can you explain more about the use of json-schema for the providers?
 * on a quick read, views.py makes sense. Longer-term, I'd feel much better about this if it were internal to the server cluster though.

manics:
Decisions from today's meeting:
- remove omero-oauth
- omero-signup should only email passwords and not display them
- provide contact information in case of problems

manics:
Screen Shot 2019-06-24 at 18 17 18

Screen Shot 2019-06-24 at 18 30 53

- Removed omero-oauth
- Changed layout, moved `If you already have account .. ` etc to below the `Create account` button. @pwalczysko If this still isn't clear can you suggest a better layout?
- If email is enabled the password is not displayed
- Optional help/contact message
- If OMERO.web has multiple servers the first is always used, it is no longer configurable

manics:
https://pub-omero.openmicroscopy.org/signup/

joshmoore:
Generally looks good. Waiting on other feedback, but I could see migrating to github/pypi as 0.1.0.

pwalczysko:
@manics Re layout https://github.com/ome/prod-playbooks/pull/178#issuecomment-505106260 - I think this is much improved. The only nitpick I have is: Maybe make it so that the email address demo-.... is not split on the dash over two lines ? Maybe just put it all on a new line ?

manics:
The help message including email address is autoformatted, I don't know how to easily prevent it being split, especially since the whole help message is configurable so we can't have a fix that only works for this particular message.

mtbc:
Had some trouble creating an account:

create

Might be unrelated to this PR though, perhaps better as issue on underlying app?

manics:
@mtbc Did you have some old browser state left behind from a previous test? There's a hidden request ID to prevent resubmission of the form if you just refresh. Retrying should work.

manics:
Unless it's a unicode issue, I ran into a problem before but couldn't work out if it's an OMERO.web function or the app that's not handling unicode input properly

mtbc:
Aha, thank you. I cleared browser state and managed to get as far as [QA 27689]( https://www.openmicroscopy.org/qa2/qa/feedback/27689/) so both your hypotheses are correct!

mtbc:
> prevent resubmission of the form

RFE: Make the error message clearer. We get the occasional person wanting demo accounts for their colleagues too so I can imagine they might try to make a few at once.

(We definitely get a fair few non-ASCII form submissions although, if it's any help, most of those would still fall within ISO 8859-1.)

pwalczysko:
>The help message including email address is autoformatted, I don't know how to easily prevent it being split, especially since the whole help message is configurable so we can't have a fix that only works for this particular message.

Understood, I think this is fine then.

manics:
Creating a user with unicode characters works in webadmin:
https://github.com/openmicroscopy/openmicroscopy/blob/v5.5.0/components/tools/OmeroWeb/omeroweb/webadmin/views.py#L468-L471
but not in the signup app:
https://gitlab.com/openmicroscopy/incubator/omero-signup/blob/v0.0.8/omero_signup/views.py#L204-209

I can't see where there's any encoding/decoding being done. Any ideas @will-moore ?

will-moore:
No, I can't either.
In ```webclient_gateway.createExperimenter()``` it simply casts to ```str()```:
```
    experimenter = ExperimenterI()
    experimenter.omeName = rstring(str(omeName))
    experimenter.firstName = rstring(str(firstName))
```

manics:
Unicode issues should be fixed now. It would be a lot easier if OMERO.py supported Python `unicode` strings.

will-moore:
That last commit doesn't *look* like if changes ```unicode``` handling?

manics:
This is the diff: https://gitlab.com/openmicroscopy/incubator/omero-signup/compare/v0.0.8...v0.0.9

I tried using `omeroweb.custom_forms.NonASCIIForm`: https://gitlab.com/openmicroscopy/incubator/omero-signup/commit/7476ebebd0107d618ca1701a8e193190e677508f
but that didn't do the validation I want- it checks for a non-empty string, but I want to specifically check for at least one alphanumeric unicode char, so I had to handle the conversion myself: https://gitlab.com/openmicroscopy/incubator/omero-signup/commit/a78d85c78a108ce68e91d2a9afb29b926774adb1

pwalczysko:
Ready to merge fmpov

will-moore:
Looks good to me and is working fine with unicode characters.

joshmoore:
From today's meeting:
 - Josh to migrate signup to github
 - Simon to rebase and redeploy on pub-omero
 - Then rolling out to demo with update to docs etc. for the next training

manics:
- Rebased
- Use release omero-signup 0.1.0
- Use a dedicated light-admin account for omero-signup

PR #181, opened 2019-07-02
NSweb default iviewer can be overridden

manics:
The default web viewer in templates/omero-web-config-for-webapps.j2 can be overridden on a case-by-case basis


joshmoore:
LGTM. The fact that we can't make use of `config_set_for_host` is unfortunate, but I assume a trickier fix.

manics:
Deployed to ns-webclients

PR #184, opened 2019-07-10
Postgres new omero db molecule workaround

manics:
Workaround https://forum.image.sc/t/fyi-not-possible-to-create-new-omero-instance-with-latest-postgresql-versions/26929
- Changes `openmicroscopy.postgresql-3` to `ome.postgresql` (existing `openmicroscopy.postgresql` is unchanged)
- Pins the pg version to `9.6.13` in the training-server molecule test. This is the only playbook that has a full OMERO molecule test.

Not yet tested

### DIFF
--- a/k8s/postgres/playbook.yml
+++ b/k8s/postgres/playbook.yml
@@ -2,7 +2,7 @@
 - hosts: ome-k8s-pg-01.openmicroscopy.org
   roles:
 
-  - role: openmicroscopy.postgresql-3
+  - role: ome.postgresql
     postgresql_version: "9.6"
     postgresql_server_listen: "'*'"
     postgresql_databases:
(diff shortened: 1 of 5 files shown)


sbesson:
Looks good. If this causes no changes against `demo` and `training` in check-mode and Travis passes, I'd say LGTM and start integrating open PRs.

Is the OMERO Dundee playbook the only one consuming the old PostgreSQL role?

manics:
Looks like it
```
$ git grep openmicroscopy.postgresql
k8s/postgres/playbook.yml:  - role: openmicroscopy.postgresql_backup
ome-dundeeomero.yml:    - role: openmicroscopy.postgresql
requirements.yml:- src: openmicroscopy.postgresql
requirements.yml:- src: openmicroscopy.postgresql_backup
```
However the existing `openmicroscopy.omero-server` role still requires `openmicroscopy.postgresql` as a transient dependency for the `psql` client so we'll need to switch to `ome.omero-server` at some point.

PR #185, opened 2019-07-11
Versioned URLs - part 2

sbesson:
Follow-up of #177. A main issue associated with the new URLs aliases (latest, minor, major) associated with every release is the absence of redirection to the canonical version URL. A typical usage is consumers relying this redirection to programmatically retrieve the latest version of a component.

This PR updates the logic for creating these aliases and creates `{latest,minor,major}/.htaccess` files redirecting to the canonical versioned URLs rather than symlinks transparently served by the Web server.

As previously, an (UoD internal-only) test of this PR has been created under http://downloads.openmicroscopy.org/omero-test/:

- the 5.4 alias has been created using the previous iteration and should serve the 5.4 content but not redirect
- the 5.5, 5 and latest aliases have been created by this PR (after manually deleting the symlinks) and should serve the 5.5 content as well as redirect the URL

The Molecule tests have also been updated to test the new logic.

### DIFF
--- a/molecule/release/prepare.yml
+++ b/molecule/release/prepare.yml
@@ -16,15 +16,30 @@
      group: root
      mode: 01555
     with_items: "{{ www_folders }}"
+  - name: Create minor version directory
+    file:
+     path: "{{ item }}/component/3.1"
+     state: directory
+    with_items: "{{ www_folders }}"
   - name: Create minor version redirects
     copy:
      dest: "{{ item }}/component/3.1/.htaccess"
      content: "Redirect 301 /component/3.1 /component/3.1.8"
     with_items: "{{ www_folders }}"
+  - name: Create major version directory
+    file:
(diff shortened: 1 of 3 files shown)


rgozim:
Looks good 👍

On a side note, last modified column however is and can be confusing when versioning is involved.

sbesson:
Thanks @rgozim, merging and will start updating the aliases. Re modification dates, I agree the Apaxy style might be one of the next things to review for improving the display of these versions.

mtbc:
https://docs.openmicroscopy.org/omero-blitz/5.5/slice2html/ doesn't seem to redirect to 
https://docs.openmicroscopy.org/omero-blitz/5.5.3/slice2html/.

sbesson:
Are these managed via these playbooks?

mtbc:
Ah, I have no idea.

PR #188, opened 2019-07-30
Replace /community with static copy

manics:
Replaces proxied phpbb `/community` with a static copy of the forums obtained using `wget --mirror -np https://www.openmicroscopy.org/community/` on 2019-07-18

Requires https://github.com/ome/ansible-role-nginx-proxy/pull/26

Deployed on https://ome-www-dev.openmicroscopy.org/community

When testing links on ome-www-dev check that no hidden requests are made to www or www-legacy domains


joshmoore:
Did a good deal of clicking around. I didn't find any hidden requests on inspection.

Things I can imagine getting rid of via parsing or CSS:
 - The "were-hiring-ribbon" 404
 - Login
 - Clicking on users names (leads to login)
 - Search
 - "Currently online" footer

In discussing, plan if all are ok with it is to get this archive "live" on www.openmicroscopy.org. At that point, google will likely still index it for findability (since search is broken without php). When/if we migrate the content to image.sc we can either delete or hide from google.

Otherwise, code changes look good. Just waiting on release versions. (Moving on to that now)

cc: @jburel @jrswedlow

sbesson:
Tested various links to the old phpBB forums from 1- other threads on the forum, 2- image.sc, 3- legacy OME mailing lists. Everything worked as expected and I did not find any hidden request.

I agree with the roadmap described in https://github.com/ome/prod-playbooks/pull/188#issuecomment-516770567. From the bullet point list, my biggest concerns are probably the search and login pages as any POST action will lead to a very bare `405` Nginx response page. Is this something we could more gracefully handle by adding extra `location` redirecting to the root website and/or Google until we hide these elements?

manics:
Last commit patches the stylesheet to hide most interactive elements (deployed to www-dev, you may need to force a refresh in your browser)

sbesson:
Now deployed live to https://www.openmicroscopy.org/community/

PR #192, opened 2019-08-08
Test ome.omero_web role with SELinux

sbesson:
--depends-on https://github.com/ome/ansible-role-omero-web/pull/21


sbesson:
Excluding as per https://github.com/ome/ansible-role-omero-web/pull/21#discussion_r311994459

sbesson:
Merging and I will deploy on `demo.openmicroscopy.org` probably over the week-end.

PR #193, opened 2019-08-09
Dropbox

pwalczysko:
Change the configuration of dropbox so that it uses regex. 
Originally, the config was the same for all users, and also, the DropBox was creating a new folder for every drop even for the same user.
This was to circumvent the (now fixed) bug with import into RO or RA group where multiple Datasets of the same name already exist.
As this bug is fixed, this PR adds more fancy regex in line with the work on the g.doc for DropBox, see last part of https://docs.google.com/document/d/1ZWBZZWHK7ZUxK48QrT_PcbVB6w72uhhtZ3atJrMD5KQ/edit#

This was run and tested on outreach.

The regex is supposed (and as tested is really doing) to do following: The name of the newly created Dataset is taken from the lowest containing folder inside the DropBox directory.

cc @manics @sbesson @jburel @joshmoore

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -355,7 +355,7 @@
     omero_server_config_set:
       #omero.fs.importUsers: "fm1"
       omero.fs.watchDir: "/home/DropBox"
-      omero.fs.importArgs: "-T Dataset:@name:from-Dropbox"
+      omero.fs.importArgs: "-T Dataset:name:from-Dropbox"
       omero.db.poolsize: 60
       omero.jvmcfg.percent.blitz: 50
       omero.jvmcfg.percent.indexer: 20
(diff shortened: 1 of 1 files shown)


jburel:
Tests
* Added a directory  to ``438CTR_01_4`` ``DropBox/trainer-1``. A dataset was successfully created and the image imported into the dataset.
 * Added a directory to ``dir-438CTR_01_4`` containing the previous directory. In that case a dataset is not created, the one created during the previous test is re-used and the image is imported.  Note that this is different that insight import workflow. Discussed with @pwalczysko , we can see both pro and cons to that approach.
* Added a directory to ``dir-2`` containing 2 directories both with images. 2 datasets are created and the images imported in the correct dataset

jburel:
Changes work as expected. Merging

PR #194, opened 2019-08-14
Remove fm users and their dropbox folders

pwalczysko:
When writing the g.doc about DropBox, I realized that we have quite a few redundant users and DropBox folders on outreach.

This PR is removing the corresponding bits from the playbook.

Also, I will remove manually the already existing DropBox folders on outreach.
This includes a "falso" ``/OMERO/DropBox`` folder, which should have never existed.

cc @jburel @joshmoore @manics @sbesson 

Edit 
Removed the redundant folder from Outreach server as indicated in the header. Dropbox on outreach still works.

### DIFF
--- a/molecule/omero-training-server/tests/test_default.py
+++ b/molecule/omero-training-server/tests/test_default.py
@@ -25,7 +25,7 @@ def test_service_running_and_enabled(Service, name):
 
 
 def test_omero_login(host):
-    with host.sudo('fm1'):
+    with host.sudo('importer1'):
         host.check_output(
             '/opt/omero/server/OMERO.server/bin/omero '
             'login -C -s localhost -u root -w omero')
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -255,15 +255,6 @@
       notify:
       - restart omero-web
 
-    - name: Add facility manager operating system users
-      become: true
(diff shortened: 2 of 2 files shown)


pwalczysko:
```
failed: [outreach.openmicroscopy.org] (item=[u'python-reportlab', u'python-markdown', u'mencoder', u'scipy', u'python-matplotlib']) => {"changed": false, "item": ["python-reportlab", "python-markdown", "mencoder", "scipy", "python-matplotlib"], "msg": "Failure talking to yum: failure: repodat
```

Not sure what to make of the travis failure - it works in real on outreach

sbesson:
The error reported above was transient and fixed by a re-run. The latest Travis error is expected since some of the Molecule tests expect a `fm1` local user - https://github.com/ome/prod-playbooks/blob/0cb3940/molecule/omero-training-server/tests/test_default.py#L28 - which is removed in this PR.

pwalczysko:
@sbesson should I try to fix the molecule test you are pointing to ? I have an idea of replacing the non-existing ``fm1`` user with ``importer1`` which does exist.

sbesson:
@pwalczysko yes if you are getting rid of `fm1`, updating the test to perform the same check using another existing user on the system should be sufficient.

PR #196, opened 2019-08-21
Update website

jburel:
Both staging and prod have been updated  with the latest tag


jburel:
I am going to update the website with the new tag so value will be changed

PR #202, opened 2019-08-30
Upgrade outreach OMERO and apps

pwalczysko:
In prep of the next week's training

cc @jburel @manics @joshmoore 

the playbook was successfully run against outreach, all worked fine.


jburel:
Connected to outreach, iviewer version is now 0.8.0

PR #204, opened 2019-09-05
Support wss external websocket connections to demo

manics:
Closes https://github.com/ome/prod-playbooks/issues/198

Deployed to pub-omero. To test connect to `wss://pub-omero.openmicroscopy.org/omero-ws`, for example using:
- https://github.com/ome/omero-py/pull/18
- https://github.com/ome/omero-insight/pull/66

This enables:
- ~~Direct `wss://host:4066` connections to OMERO using a self signed certificate~~ blocked by the firewall
- `wss://host:443/omero-ws` connections to OMERO via Nginx (port `443` is the default wss port in OMERO.py and soon the [Java Gateway](https://github.com/ome/omero-gateway-java/pull/21) so can be omitted

### DIFF
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -47,6 +47,42 @@
         - python-markdown
         - mencoder # For the 'make movie' script
 
+    - name: OMERO.server self-signed certificate directory
+      become: yes
+      file:
+        path: /opt/omero/server/selfsigned
+        state: directory
+
+    - name: OMERO.server self-signed certificate
+      become: yes
+      command: >-
+        openssl req
+        -new -nodes -x509
+        {{ omero_server_websocket_internal_cert_params }}
+        -keyout server.key
+        -out server.pem
+        -extensions v3_ca
+      args:
+        chdir: /opt/omero/server/selfsigned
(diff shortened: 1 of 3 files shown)


dominikl:
Connecting to `wss://pub-omero.openmicroscopy.org/omero-ws` works. But `wss://pub-omero.openmicroscopy.org:4066` gives 'Connection refused'. Is this expected? I.e. just internally accessible for nginx to connect to?

manics:
Sorry, I forgot it's firewalled so not accessible:
https://github.com/ome/prod-playbooks/blob/af25a297a18b09f8599f6e8cab36f3c9f3977e10/omero/omero-firewall.yml#L60
I don't see any value in opening it since the main use-case is to access omero over a standard web port.

manics:
By default Nginx doesn't verify the certificate used by by backend apps. I can switch to the wildcard cert but it means making it private key readable by OMERO instead of only by root.

sbesson:
Incidentally this has been deployed on https://demo.openmicroscopy.org/ as part of the latest execution of `site.yml`

PR #205, opened 2019-09-13
Use ome.ssl_certificate role

sbesson:
- switches from the deprecated `openmicroscopy.ssl-certificate` to `ome.ssl_certificate`
- updates playbooks to consume the new role
- defines handlers reloading the nginx server when the SSL certificate has changed

### DIFF
--- a/jupyterhub/playbook.yml
+++ b/jupyterhub/playbook.yml
@@ -206,6 +206,14 @@
       read_timeout: 86400
       host_header: "$host"
 
+  handlers:
+    - name: reload nginx
+      listen: ssl certificate changed
+      become: true
+      service:
+        name: nginx
+        state: reloaded
+
   vars:
     https_certificate_domain: outreach-analysis.openmicroscopy.org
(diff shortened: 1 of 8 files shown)


joshmoore:
Hear, hear for more renaming PRs.

sbesson:
Last commit  defines handlers restarting nginx and was sucessfully tested against `ome-www-dev`. Once this is reviewed and merged, I expect running `site.yml` should be sufficient to update SSL certificates.

jupyterhub/playbook.yml
```diff
@@ -206,6 +206,14 @@
       read_timeout: 86400
       host_header: "$host"
 
+  handlers:
```
manics: This server is currently using certbot to automatically renew lets-encrypt certificates. The `ssl-certificate` role is only used to generate self-signed certificates for testing. There's no harm having this here, but in practice it'll never be used unless we get rid of lets-encrypt on the training/outreach servers.

PR #211, opened 2019-10-28
Python3

sbesson:
Travis has been failing recently on this repository with the same failure as the one that https://github.com/ome/ansible-roles/issues/9 is trying to address.

To fix it, this PR uncaps the Python version on Travis CI (effectively migrating to Python 3) and fixes a few Python tests accordingly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@
 sudo: required
 language: python
 
-services:
-  - docker
-
 install:
   - pip install "ome-ansible-molecule<0.5"
   - pip install jmespath
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 ---
 sudo: required
 language: python
-python: 2.7
 
 services:
   - docker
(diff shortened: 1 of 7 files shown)


sbesson:
From https://travis-ci.org/ome/prod-playbooks/builds/603851390?utm_medium=notification&utm_source=github_status, all matrix jobs are now passing except for https://travis-ci.org/ome/prod-playbooks/jobs/603851403 which fails due to timeout.

I suspect this is an issue with one of the dependencies in the stack (Python, Molecule, pytest, testinfra) so I will probably try going through issues/release notes. Any hinsights from @ome/devops welcome.

manics:
Does the failing scenario pass locally?

sbesson:
Locally (after reducing `requirements.yml` to prevent failures in the `dependency` step), all `www` tests systematically fail with a `Too many open files`. Commenting `test_internal_redirects` suffices to let them pass. Looking into the size of `ulimit` on the Ansible-Docker image.

manics:
Ulimits are set on the Docker host: https://github.com/ansible/molecule/issues/1091

sbesson:
Local testing confirmed that the `host.check_output()` pytests are slow (hence causing Travis CI timeouts) and additionally causing `Too many open files` error for low `ulimit` values.

I tried various alternatives including upgrading `testinfra` up to 3.2.0 (https://travis-ci.org/sbesson/prod-playbooks/builds/604176419) or looping over the `check_output` calls within the tests. None of these reliably improved the testing speed both locally and on Travis. Last two commits propose to skip these tests for now to prevent the failure of this repository.

PR #213, opened 2019-10-30
Update public image folders to consume new public OME-TIFF/OME-XML folders

sbesson:
As discussed with @ome/formats, the internal name of the public folders has been changed to `public` to match the other formats.

This playbook has been deployed so that https://downloads.openmicroscopy.org/images/OME-TIFF/ and https://downloads.openmicroscopy.org/images/OME-XML/ keep working.

Proposing to merge as soon once Travis is green.


sbesson:
Merging as per the description to keep the playbook in sync with the production state.

PR #214, opened 2019-11-01
Upgrade figure

pwalczysko:
Upgrade in anticipation of the coming workshop. This was successfully run both against outreach and workshop servers.

cc @sbesson @manics @jburel


jburel:
Checked both outreach and workshop
Figure 4.1.0 has been installed

PR #215, opened 2019-11-12
Outreach Python 3

manics:
Optionally switch to Python 3 in outreach


manics:
See https://github.com/ome/prod-playbooks/pull/225

PR #218, opened 2019-12-09
Force upgrade of OMERO.iviewer on Virtual Microscope.

mtbc:
OMERO.iviewer on https://learning.openmicroscopy.org/ is still back on v0.7.1. This PR ensures that when the playbook is run for the next scheduled maintenance of Virtual Microscope OMERO.iviewer will be upgraded to the latest stable version.

### DIFF
--- a/learning.yml
+++ b/learning.yml
@@ -113,7 +113,7 @@
     - role: ome.omero_web_apps
       omero_web_apps_packages:
         - omero-gallery
-        - omero-iviewer>=0.8
+        - omero-iviewer==0.8.1
         - omero-virtual-microscope
 
   tasks:
--- a/learning.yml
+++ b/learning.yml
@@ -112,9 +112,9 @@
 
     - role: ome.omero_web_apps
       omero_web_apps_packages:
-        - omero-gallery
+        - omero-gallery==3.1.1
         - omero-iviewer==0.8.1
-        - omero-virtual-microscope
+        - omero-virtual-microscope==1.0.1
 
   tasks:
     - name: TLS certificate is installed for JVM
--- a/learning.yml
+++ b/learning.yml
@@ -40,7 +40,7 @@
     - role: ome.omero_server
       ice_python_wheel: https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.1.0/zeroc_ice-3.6.4-cp27-cp27mu-linux_x86_64.whl
       omero_server_systemd_start: False
-      omero_server_release: latest
+      omero_server_release: 5.5.1
       omero_server_dbpassword: "{{ database_password }}"
       omero_server_rootpassword: "{{ server_password }}"
       omero_server_config_set:
@@ -77,7 +77,7 @@
       ice_python_wheel: https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.1.0/zeroc_ice-3.6.4-cp27-cp27mu-linux_x86_64.whl
       omero_web_setup_nginx: False
       omero_web_systemd_start: False
-      omero_web_release: latest
+      omero_web_release: 5.5.1
       omero_web_config_set:
         omero.web.server_list:
           - ["localhost", 4064, "Virtual Microscope"]
--- a/learning.yml
+++ b/learning.yml
@@ -113,7 +113,7 @@
     - role: ome.omero_web_apps
       omero_web_apps_packages:
         - omero-gallery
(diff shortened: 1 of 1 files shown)


learning.yml
```diff
@@ -113,7 +113,7 @@
     - role: ome.omero_web_apps
       omero_web_apps_packages:
         - omero-gallery
-        - omero-iviewer
+        - omero-iviewer>=0.8
```
sbesson: Main downside of this form is that it makes the playbook non-idempotent. Have you considered pinning it to the latest stable release and managing the app versions via PRs ?

mtbc: I may not understand what you mean by idempotence. Surely it'll have the same effect whether I run it for-real once or thrice but I may be mistaken?

Certainly considered it: Definitely would like to avoid fixing to particular version because that's going to go out of date very quickly. E.g., if we have an iviewer 0.8.2 out soon (will we?) then we have to go through extra PR churn just to make sure I get that in instead next week. Similarly the playbook installs a whole lot of other stuff too (Python and otherwise) without fixing version.

Could be that we have something somewhere that fixes desired production versions in group vars for a ton of stuff and lots of playbooks could use that -- maybe we already do? -- but I think it'll be more pain than benefit to do per-playbook.

Probably all part of a wider discussion I've been only peripheral for. If you'd like me to fix to 0.8.1 I certainly can.

mtbc: (Or I could also specify `<0.9` or something if that might seem safer.)

manics: Idempotence: Unless otherwise stated It should be possible for any admin to run any playbook at any point without it changing a system. If a dependency is open-ended then when there is a new release it'll get upgraded when the playbook is run.

mtbc: Oh! I thought that `>=0.8` would leave it at `0.8.1` if that is installed already but now `0.9` is out. If that's not so will definitely have to look back over what else I've done.

manics: Actually you're right, it will, but if the system has to be reinstalled from scratch for disaster recovery you could end up with a different plugin version

mtbc: This is true of a bunch of what else gets installed too though, right? Right from the early `yum install ...` stuff onward.

manics: Yes, but the official RHEL yum repos should guarantee compatibility/stability between updates of RPMs. `>=0.8` doesn't guarantee this, though including `<0.9` should. Depends how much control you want over the system.

mtbc: I'd worry that in the longer term `<0.9` could *cause* incompatibility for a fresh installation without also pinning versions of `omero-web`, etc. and if limiting upgrades widely that's a lot among which to look out for security vulnerability announcements (I'd guess that's why we have simply `gunicorn>=19.3`), hence probably good to move to group vars file we can widely share. Will test the `<0.9` here though to get this PR in! :smiley:

manics: I missed that- OMERO.server and OMERO.web versions should also be pinned for a production system unless you want unexpected upgrades.

manics: Do you want to fix that in this PR or a separate one?

mtbc: I thought that for `omero_server_release` and `omero_web_release` the default is simply `present` so running the playbook won't cause an unexpected upgrade? Or is it another facet of the issue with what if the VMware snapshots fail so we need a new Virtual Microscope and we pin things in case latest iviewer release doesn't work with latest server release or suchlike? (Not inconceivable with the need for things like this Python 3 effort now.)

Should I just put numbers in for those variables matching git tags or is it better to use these `*_override` variables instead? Can certainly do so in this PR.

manics: `present` wouldn't be so bad, though it still has the same problem of what happens with a redeployment. It doesn't matter though, it's currently set to `latest` 😀:
https://github.com/ome/prod-playbooks/blob/3a3739174e88dba5798ddeacf8ff67e893b6221d/learning.yml#L43

You can use whatever style works for you, either have `_override` variables or move `omero_*_release` to the playbook level vars section so it's clearer than having the vars nested inside the role.

mtbc: Ah, good idea, thank you: I'll move them to a `vars` section to make it easier.

mtbc: Will move to vars section in followup to avoid conflicts with https://github.com/ome/prod-playbooks/pull/219.

manics:
Can be merged when it's ready to deploy.

mtbc:

PR #219, opened 2019-12-09
start services on boot: learning, sls-repo

mtbc:
Both `learning.yml` and `sls-gallery.yml` include,
```
omero_server_systemd_start: False
omero_web_systemd_start: False
omero_web_setup_nginx: False
```
It wasn't clear to me that this also prevents the services from being enabled at all. This extra task fixes that issue. The services are already otherwise enabled on the actual server but this PR reflects that.

### DIFF
--- a/learning.yml
+++ b/learning.yml
@@ -133,3 +133,13 @@
         src: "omero/files/learning-omero-web.conf"
         dest: "/etc/nginx/conf.d/omero-web.conf"
       notify: restart nginx
+
+    - name: services are started on boot
+      become: yes
+      service:
+        name: "{{ item }}.service"
+        enabled: true
+      loop:
+        - nginx
+        - omero-server
+        - omero-web
--- a/learning.yml
+++ b/learning.yml
@@ -39,7 +39,6 @@
 
     - role: ome.omero_server
       ice_python_wheel: https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.1.0/zeroc_ice-3.6.4-cp27-cp27mu-linux_x86_64.whl
-      omero_server_systemd_start: False
       omero_server_release: latest
       omero_server_dbpassword: "{{ database_password }}"
       omero_server_rootpassword: "{{ server_password }}"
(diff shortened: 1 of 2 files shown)


mtbc:
It's possible that there's some way to write the playbooks such that now on a fresh install they'll run through okay without those `False` lines in the first place but it hardly seems high priority to figure that out given that the servers would then just be restarted more.

manics:
Why were the values set to `False` in the first place?

mtbc:
The playbook would fail to run through a fresh install, I guess something important the servers needed wasn't set yet. Probably fixable, just a case of when to reopen that can.

mtbc:
I forget what was going on with server but I wonder if web was doing "too much" setup with my having a custom nginx conf, etc.

mtbc:
Outstanding issue looks to be (at least) web app installation which in the longer term we can figure out.

snoopycrimecop:
Conflicting PR. Removed from build [TRAINING-merge#432](https://ci.openmicroscopy.org/job/TRAINING-merge/432/). See the [console output](https://ci.openmicroscopy.org/job/TRAINING-merge/432/consoleText) for more details.
Possible conflicts:
  - PR #218 mtbc 'Force upgrade of OMERO.iviewer on Virtual Microscope.'
    - learning.yml

--conflicts

mtbc:
Merge #218 in since latest push there conflicted here.

learning.yml
```diff
@@ -112,9 +111,9 @@
 
     - role: ome.omero_web_apps
       omero_web_apps_packages:
-        - omero-gallery
-        - omero-iviewer
-        - omero-virtual-microscope
+        - omero-gallery==3.1.1
+        - omero-iviewer==0.8.1
+        - omero-virtual-microscope==1.0.1
```
jburel: Not for this PR
I don't think we look at ``omero-virtual-microscope`` in the context of py3
Have you checked @mtbc?

mtbc: This is from the extra commits needed to get the other PR approved that then conflicted with this one so I merged it in.

I made sure that `omero-virtual-microscope` on the gdoc spreadsheet of apps. I don't yet know what it needs before it will work.

mtbc:

PR #220, opened 2019-12-12
separate /opt/omero/ mount on learning and sls-repo

mtbc:
Adds space for OMERO.server and OMERO.web to live separately on `/opt/omero/`. Should already pass for SLS Gallery but will fail on Virtual Microscope until existing data is migrated during the maintenance window.


mtbc:

PR #223, opened 2019-12-19
Gather current release versions into global variables.

mtbc:
Gathers the various release-version values together from the Virtual Microscope and SLS Gallery playbooks, replacing any "latest" with the currently installed version.

### DIFF
--- a/learning.yml
+++ b/learning.yml
@@ -46,8 +46,6 @@
       lvm_shrink: False
 
     - role: ome.omero_server
-      ice_python_wheel: https://github.com/ome/zeroc-ice-py-centos7/releases/download/0.1.0/zeroc_ice-3.6.4-cp27-cp27mu-linux_x86_64.whl
-      omero_server_release: 5.5.1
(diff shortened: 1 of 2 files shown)


mtbc:
Once this is in I'll open PRs for upgrade to OMERO 5.6.

learning.yml
```diff
@@ -119,9 +115,9 @@
 
     - role: ome.omero_web_apps
       omero_web_apps_packages:
-        - omero-gallery==3.1.1
-        - omero-iviewer==0.8.1
-        - omero-virtual-microscope==1.0.1
+        - omero-gallery=={{ omero_web_apps_release["omero_gallery"] }}
```
manics: FYI if you have plain variable names you can also use `{{ omero_web_apps_release.omero_gallery }}`

joshmoore: Leaving that to the follow up.

mtbc: :+1:

PR #224, opened 2020-01-07
Compress OMERO.server logs for learning and sls-repo.

mtbc:
The Virtual Microscope and SLS Gallery servers have finite space for `/opt/omero/` and OMERO.server logs are so large and compressible that it seems well worth saving that space so this PR adjusts the Logback configuration to compress log files on rollover.

This has already been run on SLS Gallery. For Virtual Microscope we'll wait for the next maintenance window such as when 5.6.1 is released.

At a quieter time we should look at where we can roll the functionality of these various tasks into existing or new roles.

### DIFF
--- a/learning.yml
+++ b/learning.yml
@@ -124,6 +124,22 @@
         - omero-virtual-microscope==1.0.1
 
   tasks:
+    - name: find OMERO.server log configuration
+      become: yes
+      find:
+        paths: /opt/omero/server/OMERO.server/etc/
+        patterns: "logback*.xml"
+      register: logbacks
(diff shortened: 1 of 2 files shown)


joshmoore:
Just a side comment that this is also doable directly from the logbook configuration. I thought I had a branch somewhere, but it didn't show up via a quick scan.

joshmoore:
Probably makes sense to add this to ansible-role-omero-server eventually if we don't find&resurrent the option in logbock.xml itself to turn on this functionality. :+1:

PR #225, opened 2020-01-07
Python3 follow-up

pwalczysko:
this is a follow-up PR on https://github.com/ome/prod-playbooks/pull/215 as discussed with @manics

PR #228, opened 2020-01-15
OMERO 5.6 for SLS Gallery and Virtual Microscope

mtbc:
Assumes the requirements updates from #225.


joshmoore:
Based on the conversations I was around for, this matches my expectations. Merging since this has been deployed to production.

PR #231, opened 2020-01-16
Add websocket connection motifs for outreach servers

pwalczysko:
After discussion with @jburel and @manics , this is motivated by the need to have a websocket connection to outreach servers for the next outreach session, to be able to use MyBinder.

This PR basically leans on https://github.com/ome/prod-playbooks/pull/204, trying to make similar changes for the outreach playbook as were done in https://github.com/ome/prod-playbooks/pull/204 for demo server.

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -58,6 +58,43 @@
       notify:
       - restart omero-web
 
+    - name: OMERO.server self-signed certificate directory
+      become: yes
+      file:
+        path: /opt/omero/server/selfsigned
+        state: directory
+
+    - name: OMERO.server self-signed certificate
+      become: yes
+      command: >-
+        openssl req
+        -new -nodes -x509
+        {{ omero_server_websocket_internal_cert_params }}
+        -keyout server.key
+        -out server.pem
+        -extensions v3_ca
+      args:
+        chdir: /opt/omero/server/selfsigned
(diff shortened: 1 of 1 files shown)


pwalczysko:
Successfully deployed on ome-training-4. Connection test is passing 2x

omero/training-server/playbook.yml
```diff
@@ -58,6 +54,67 @@
...
+        -keyout server.key
+        -out server.pem
+        -extensions v3_ca
+      args:
+        chdir: /opt/omero/server/selfsigned
+        creates: server.pem
+
+    # This self-signed cert is used to encrypt the websocket connection
+    # between Nginx and OMERO.server
+    - name: OMERO.server self-signed certificate pkcs12
```
mtbc: I'm guessing this isn't the kind of thing that `ome.ssl_certificate` can yet do for us.

jburel:
I can connect to outreach and workshop using ``ws://workshop.openmicroscopy.org/omero-ws`` via insight but I cannot via omero-py
I have tried using directly the ``BlitzGateway`` and ``omero.client``

jburel:
ok 
so the problem was due to the way we specify
java and python are handling things differently
wss required for python

jburel:
Merging to that @pwalczysko can continue the upgrade.

PR #236, opened 2020-02-04
website update

jburel:
update version and sha256

Updated prod and staging


sbesson:
I assume this should have been merged long ago to avoid the website to be downgraded. Doing so now to bump to another version of the static bundle

PR #237, opened 2020-02-17
Use omero.web.nginx_server_extra_config instead of patching generated config

manics:
This removes on of the idempotence errors by declaritively setting the extra nginx config isntead of patching it.

Already deployed on demo since it was being restarted anyway whilst investigating a problem (still not working though)


mtbc:
Only question is the Travis failure really, probably unrelated?

manics:
Restarted the failing job, now green

mtbc:
:+1:

PR #243, opened 2020-03-25
demoserver_reportlab

will-moore:
See https://forum.image.sc/t/error-when-exporting-pdf-from-omero-figure/35460

This should fix the install of reportlab on demo server. Uses the commands copied from https://github.com/ome/prod-playbooks/blob/5a9711b718233f3583811e3dfe4ce67a5fb28ee6/ome-dundeeomero.yml#L295

### DIFF
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -36,15 +36,12 @@
         name: mencoder
         state: present
 
-    - name: OMERO.figure server-side prerequisites, script prerequisites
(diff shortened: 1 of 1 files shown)


mtbc:
pub-omero could be useful for testing this.

sbesson:
Now deployed on `demo.openmicrosscopy.org`. Leaving you to double check everything is working as expected and we can probably update the public thread.

PR #244, opened 2020-03-25
Add public downloads folder for OBF file format

sbesson:
See https://github.com/ome/bioformats/pull/3528#issuecomment-600219568


sbesson:
Got :+1: by @melissalinkert so moving forward with the deployment of this in production

PR #245, opened 2020-04-14
Add ome.sudoers to www playbook

sbesson:
This should allow additional AD groups to get access to this server

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -96,7 +96,7 @@
   version: 0.3.2
 
 - src: ome.sudoers
-  version: 1.0.1
+  version: 1.0.2
 
 - src: ome.system_monitor_agent
   version: 0.1.1
--- a/www/playbook.yml
+++ b/www/playbook.yml
@@ -59,6 +59,8 @@
       tags: monitoring
       when: "'10.1.255.216' in ansible_dns.nameservers"
 
+    - role: ome.sudoers
+
   post_tasks:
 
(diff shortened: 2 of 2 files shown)


sbesson:
Actually I had to patch the role inline with `s|visudo|/usr/sbin/visudo` in order for the command to complete. I'll open a PR shortly (trying to construct failing tests) but does that ring a bell to you?

manics:
No, it's always worked fine for me, e.g. all the new VMs that used the bootstrap playbook worked https://github.com/ome/prod-playbooks/blob/master/bootstrap/playbook.yml

sbesson:
I have not been able to identify why this server would be different (`RHEL7` vs `CentOS7`?). Opened https://github.com/ome/ansible-role-sudoers/pull/1 with the associate fix used for deploying this change in production.

Leaving the reviewer of this PR to check whether this change is necessary to run the playbook against e.g. the `ome-www-dev` server. If still necessary, I propose to also review/release the role PR and bump the role dependency  before merging.

manics:
I needed https://github.com/ome/ansible-role-sudoers/pull/1 for `ome-www-dev`, I don't know what's different about that server.

sbesson:
@manics thanks for confirming independently. I cut a patch of `ome.sudoers` with the `visudo` fix and bumped the required version in 16f52ed .

manics:
Thanks!

PR #251, opened 2020-05-05
Add public samples images for MetaXpress

sbesson:
See also https://github.com/ome/bio-formats-documentation/pull/157

This should create a public folder under https://downloads.openmicroscopy.org/images/ for the public MetaXpress filesets in our curated QA repository /cc @melissalinkert


sbesson:
No objection was raised at the Formats meeting.

PR #255, opened 2020-05-14
training-server: activate omero-ms-zarr

joshmoore:
This starts the latest docker image for omero-ms-zarr from Docker Hub, listening on port 8080. This will allow us to point notebooks and similar at:

```https://workshop.o.org:8080/image/$ID.zarr```

Suggestions for similar other configuration welcome. It is unclear how many concurrent users this single endpoint will be able to handle.

cc: @pwalczysko @mtbc @will-moore @jburel @manics

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -345,6 +345,8 @@
           -A POSTROUTING -o {{ external_nic }} -j MASQUERADE
           # Allow world to access 10389?
           -A INPUT -p tcp -m tcp --dport 10389 -s 0.0.0.0/0 -j ACCEPT
+          # Allow world to access 8080?
+          -A INPUT -p tcp -m tcp --dport 8080 -s 0.0.0.0/0 -j ACCEPT
         state: present
 
   # Crypted passwords generated using
@@ -387,7 +389,7 @@
     # or override it by defining the private variable os_system_users_password_override.
     os_system_users_password: "{{ os_system_users_password_override | default('$6$leKi5B1PgSvdA/ec$xbU3CnoSFnYdeZjEjKK5TH8SGATsW746uopssff4edpgyu.cWXGo9A.oK6wH9kIkxLCCNcORGZnnroZPMqGzN/') }}"
     apache_docker_release: "{{ apache_docker_release_override | default('0.6.0') }}"
-    omero_ms_zarr_release: "{{ omero_ms_zarr_release | default('latest') }}"
+    omero_ms_zarr_release: "{{ omero_ms_zarr_release_override | default('latest') }}"
     ldap_password: "{{ ldap_password_override | default ('secret') }}"
     omero_server_config_set:
       #omero.fs.importUsers: "fm1"
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -319,8 +319,10 @@
         network_mode: host
         published_ports:
           - "8080:8080"
-        state: started
         restart_policy: always
+        state: started
+        volumes:
+          - "/OMERO:/OMERO:ro"
 
     # https://fralef.me/docker-and-iptables.html
     # https://blog.daknob.net/debian-firewall-docker/
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -114,13 +114,6 @@
         dest: /etc/nginx/conf.d/websockets.conf
       # Don't notify, nginx isn't installed yet
 
-    - name: NGINX - omero-ms-zarr support
-      become: yes
-      template:
-        src: ../../templates/nginx-confdnestedincludes-omero-ms-zarr.j2
-        dest: /etc/nginx/conf.d-nested-includes/omero-ms-zarr.conf
-      # Don't notify, nginx isn't installed yet
-
   roles:
 
     - role: ome.postgresql
@@ -323,12 +316,19 @@
           CONFIG_omero_db_name: omero
           CONFIG_omero_data_dir: /OMERO
         network_mode: host
-        exposed_ports: "8080"
-        published_ports: "8080:8080"
         restart_policy: always
         state: started
+        pull: yes
         volumes:
           - "/OMERO:/OMERO:ro"
+
+    - name: NGINX - omero-ms-zarr support
+      become: yes
+      template:
+        src: ../../templates/nginx-confdnestedincludes-omero-ms-zarr.j2
+        dest: /etc/nginx/conf.d-nested-includes/omero-ms-zarr.conf
+      notify:
+        - restart nginx
 
     # https://fralef.me/docker-and-iptables.html
     # https://blog.daknob.net/debian-firewall-docker/
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -114,6 +114,12 @@
         dest: /etc/nginx/conf.d/websockets.conf
       # Don't notify, nginx isn't installed yet
 
+    - name: NGINX - omero-ms-zarr support
+      become: yes
+      template:
+        src: ../../templates/nginx-confdnestedincludes-omero-ms-zarr.j2
+        dest: /etc/nginx/conf.d-nested-includes/omero-ms-zarr.conf
+      # Don't notify, nginx isn't installed yet
 
   roles:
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jburel** [comment]
I think I will use that approach in the ilastik notebook instead of ``endpoint_url = 'https://minio-dev.openmicroscopy.org/'`` 
since we can then show both use cases in action cc @dominikl

**Comment by joshmoore** [comment]
One potential nice to have would be to set add a read-only DB user to the system.

**Comment by joshmoore** [comment]
A check-mode run shows:

```
TASK [Run docker for omero-ms-zarr] ********************************************************************************************************************************************************************************************
changed: [workshop.openmicroscopy.org]

TASK [Iptables Docker forward rules] *******************************************************************************************************************************************************************************************
System rules were not changed (e.g. rule weight changed, redundant rule, etc)
changed: [workshop.openmicroscopy.org]

TASK [Iptables Docker nat rules] ***********************************************************************************************************************************************************************************************
--- before
+++ after
@@ -1,4 +1,5 @@
 *nat
 -A POSTROUTING -o ens192 -j MASQUERADE -m comment --comment ansible[docker_outreach_nat]
 -A INPUT -p tcp -m tcp --dport 10389 -s 0.0.0.0/0 -j ACCEPT -m comment --comment ansible[docker_outreach_nat]
+-A INPUT -p tcp -m tcp --dport 8080 -s 0.0.0.0/0 -j ACCEPT -m comment --comment ansible[docker_outreach_nat]
 COMMIT

changed: [workshop.openmicroscopy.org]
```

however I'm currently having issues starting the image manually on workshop.

**Comment by joshmoore** [comment]
@pwalczysko _et al_. this has been updated to match workshop.o.org. I haven't deployed to any other servers.

**Comment by pwalczysko** [comment]
>....updated to match workshop.o.org. 

This assumes you somehow did something on workshop manually ?
And you mean you did not run the playbook against any servers ?

**Comment by joshmoore** [comment]
> This assumes you somehow did something on workshop manually ?

I tested the solution manually, but then I ran the playbook against just workshop and not the other servers in the inventory using `-l workshop.openmicroscopy.org`

**Comment by mtbc** [comment]
Diff certainly looks plausible.

File: omero/training-server/playbook.yml [hunk 1]
```diff
@@ -305,6 +311,25 @@
...
+        image: openmicroscopy/omero-ms-zarr:{{ omero_ms_zarr_release }}
+        name: omero_ms_zarr
+        env:
+          CONFIG_omero_db_host: localhost
+          CONFIG_omero_db_user: omero
+          CONFIG_omero_db_pass: omero
+          CONFIG_omero_db_name: omero
+          CONFIG_omero_data_dir: /OMERO
+        network_mode: host
+        exposed_ports: "8080"
```
manics, line 326:
```suggestion
```
You don't need this since it's included in `published_ports`, and in any case it's [already in the Dockerfile](https://github.com/ome/omero-ms-zarr/blob/364057c77f7c079907653fd9d03bb41504ef9a9d/Dockerfile#L19).

omero/training-server/playbook.yml
```diff
@@ -305,6 +311,25 @@
...
+        published_ports: "8080:8080"
```
manics, line 327:
```suggestion
```
[This doesn't do anything](https://docs.docker.com/network/host/) since you're running with `network_mode: host`

omero/training-server/playbook.yml
```diff
@@ -371,6 +396,7 @@
     # or override it by defining the private variable os_system_users_password_override.
     os_system_users_password: "{{ os_system_users_password_override | default('$6$leKi5B1PgSvdA/ec$xbU3CnoSFnYdeZjEjKK5TH8SGATsW746uopssff4edpgyu.cWXGo9A.oK6wH9kIkxLCCNcORGZnnroZPMqGzN/') }}"
     apache_docker_release: "{{ apache_docker_release_override | default('0.6.0') }}"
+    omero_ms_zarr_release: "{{ omero_ms_zarr_release_override | default('latest') }}"
```
manics, line 399: Is this going to be tagged? If not you may want to set [`pull`](https://docs.ansible.com/ansible/latest/modules/docker_container_module.html#parameter-pull) to ensure the latest image is always pulled.

templates/nginx-confdnestedincludes-omero-ms-zarr.j2
```diff
@@ -0,0 +1,7 @@
+location /image {
+    proxy_pass http://127.0.0.1:8080/image;
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection $connection_upgrade;
+    proxy_read_timeout 86400;
```
manics, line 6:
```suggestion
```
You shouldn't need these, they're only required for websockets.

omero/training-server/playbook.yml
```diff
@@ -114,6 +114,12 @@
         dest: /etc/nginx/conf.d/websockets.conf
       # Don't notify, nginx isn't installed yet
 
+    - name: NGINX - omero-ms-zarr support
+      become: yes
+      template:
+        src: ../../templates/nginx-confdnestedincludes-omero-ms-zarr.j2
+        dest: /etc/nginx/conf.d-nested-includes/omero-ms-zarr.conf
+      # Don't notify, nginx isn't installed yet
```
manics, line 122: As you probably guessed from the comment if this is run on an existing server any change to this file will require a manual restart of `nginx`.

If you move this to the `tasks:` section you should be able to notify the nginx restart handler. Alternatively you could append it to
https://github.com/ome/prod-playbooks/blob/eb60dbda1082b02e18aafafbfaabae94761293eb/nightshade-webclients.yml#L107-L108

Arguably the preceding `NGINX - websocket proxy support` config should go in there too but that's a separate issue

omero/training-server/playbook.yml
joshmoore: These were all leftovers from trying to get iptables to work. Do you have any idea what went wrong there? (Several docker threads showed needing to explicitly list exposed and published ports for `docker_container`, but since I never got it working, I don't know)

omero/training-server/playbook.yml
joshmoore: Where would you set `pull`? (Generally, yes it will be tagged, but I assume a few more PRs before we're ready for the meeting)

omero/training-server/playbook.yml
manics: By default Docker creates its own Iptables rules, which interacts with the rules creates by the iptables_raw module. There are several unsatisfactory ways of getting around this, the option used in this playbook is to prevent Docker updating the iptables rules:
https://github.com/ome/prod-playbooks/blob/13a7c298439be9aff940c7c7fd4c934cbc793a4e/omero/training-server/playbook.yml#L149-L152

Best way to add a forwarding rule is probably to register the output of the `docker_container` module which should include the IP.

omero/training-server/playbook.yml
manics: In the `docker_container` task.

joshmoore:
@manics: pushed

joshmoore:
Thanks. @pwalczysko : a re-run on workshop would be good to deploy v0.1.2

jburel:
Using omero-ms-zarr did not work when pointing to workshop this morning
i.e. ``workshop.o.org:8080/image/id.zarr``

joshmoore:
The iptables solution didn't work, so we're proxying through https:// port 443:

```
ome_zarr info https://workshop.openmicroscopy.org/image/1.zarr
root_attrs {'multiscales': [{'datasets': [{'path': '0'}], 'version': '0.1'}], 'omero': {'channels': [{'color': '00FF00', 'coefficient': 1.0, 'active': True, 'label': '525', 'window': {'min': 90.0, 'max': 367.0, 'start': 103.85, 'end': 218.11249999999998}, 'family': 'linear', 'inverted': False}, {'color': 'FF0000', 'coefficient': 1.0, 'active': True, 'label': '632', 'window': {'min': 89.0, 'max': 151.0, 'start': 89.0, 'end': 151.0}, 'family': 'linear', 'inverted': False}], 'rdefs': {'defaultT': 96, 'model': 'color', 'defaultZ': 0}, 'name': 'test11_R3D.dv', 'id': 1}}
```

manics:
Even if it did work port 8080 is blocked at the UoD firewall. Are there any advantages in *not* proxying through Nginx?

pwalczysko:
All the commits from here taken and deployed on wrorkshop and outreach servers.
Went fine, the servers are alive and browsing and viewing is intact.

cc @joshmoore

PR #273, opened 2020-07-20
Demo 5.6.2 upgrade

sbesson:
Prepares the upgrade of the demo servers to the latest released OMERO components - see https://www.openmicroscopy.org/2020/07/16/omero-5-6-2.html

As discussed earlier with @mtbc @manics @will-moore, also includes an upgrade of the OMERO Ansible roles to their latest versions including the updates to the `ome.postgresql` variables across all production playbooks.

### DIFF
--- a/learning.yml
+++ b/learning.yml
@@ -21,7 +21,6 @@
       lvm_shrink: False
 
     - role: ome.postgresql
-      postgresql_install_server: True
       postgresql_databases:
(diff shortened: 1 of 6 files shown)


sbesson:
`pub-omero/demo` now upgraded with these changes. Proposing to merge so that upcoming server/web upgrades can consume the requirements changes.

PR #275, opened 2020-07-21
SLS Gallery upgrade.

mtbc:
Now deployed.

Curiously, had to restart omero-web manually for the upgraded version to be running.


sbesson:
The `omero-web` restart issue was captured as #276

PR #279, opened 2020-08-24
Bump PostgreSQL version to 11 on OME demo server playbook

sbesson:
As discussed this morning, this will manually require a dump of the DB, an installation + uninstallation of the current PSQL and a restore of the DB.

To be tested on `pub-omero` first

### DIFF
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -383,7 +383,7 @@
     omero_cli_render_release: "{{ omero_cli_render_release_override | default('0.5.0') }}"
     omero_metadata_release: "{{ omero_cli_metadata_release_override | default('0.5.0') }}"
 
-    postgresql_version: "9.6"
+    postgresql_version: "11"
     filesystem: "xfs"
 
     omero_server_config_set:
--- a/requirements.yml
+++ b/requirements.yml
@@ -61,6 +61,9 @@
 - src: ome.omero_web
   version: 3.1.0
 
+- src: ome.python3_virtualenv
+  version: 0.1.1
+
 # Delete this when everything is on Python 3
 - src: ome.omero_web_apps
   version: 0.3.0


sbesson:
Successfully deployed on `pub-omero`. A few notes on the upgrade process:

- stopped the server manually using `service omero-server stop`
- dumped the DB using `pg_dump -h localhost -U demo -Fc -f omero-20200824.pg_dump omero`
- deleted the PSQL 9.6 packages using `sudo yum remove postgresql*`
- installed PSQL 11 by running the Ansible playbook with this PR included up to `ome.postgresql`
- restored the DB using `pg_restore -h localhost -U demo -Fc -d omero omero-20200824.pg_dump` (note the 9.6 data under `/var/lib/pgsql` is _not_ uninstalled so I removed it manually as the partition was reaching 100% of usage)
- re-run the playbook to completion which restarted the server

Overall, it worked and the DB is happening running on PSQL 11 but the mixture of local commands/playbooks gave me a fairly uneasy feeling of potentially leaving the system in a half-broken state.

Happy to move forward and run the same process on `demo.openmicroscopy.org` or test an alternative procedure on the staging system if someone thinks of something. /cc @manics @joshmoore

manics:
Upgrading `demo` as suggested sounds fine to me.

`pg_upgrade --link` https://www.postgresql.org/docs/11/pgupgrade.html sounds interesting (and new?) but would require testing, and the infrequency of PG upgrades means it's probably not worth spending too much time on this.

sbesson:
`demo.openmicroscopy.org` now upgraded

joshmoore:
Generally looks good. Primary question is now nightshade downtime?

sbesson:
Definitely, something that could be applied to nightshade as well (will require a variable bump in a separate playbook.

From my experience with demo, the downtime window was fairly small, the major bottleneck being the DB restoration. Given the size of the DB on nightshade, I would roughly plan for 2h.

manics (approved):
Is there a pressing reason to upgrade Postgres on Nightshade now? If not could we leave it for the next OMERO.server release?

joshmoore:
Primary reason would be to gain more confidence in the pgdata directory.

PR #280, opened 2020-09-09
Omero firewall update

manics:
Changes the GPFS and check-mk source IPs to be configurable since they're useless to anyone outside who uses these playbooks as a reference.


joshmoore (approved):
LGTM

PR #285, opened 2020-09-16
Replace nightshade pg cronjob with ome.postgresql_backup

manics:
Recently we discovered the current pgdump directory format may lead to loss of the previous backup if an error occurs during the current backup.

This replaces the custom cron job with the [ome.postgresql_backup role](https://github.com/ome/ansible-role-postgresql-backup) as used in https://github.com/ome/prod-playbooks/blob/2afc0764391e467411745946b5577973d7fa5625/k8s/postgres/playbook.yml#L25-L28

Note this role dumps all databases

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -74,9 +74,8 @@
 - src: ome.postgresql
   version: 5.0.1
 
-- name: ome.postgresql_backup
-  src: https://github.com/ome/ansible-role-postgresql-backup/archive/acd03814b8b5dd891451407b28820bfbd4fadbc5.tar.gz
-  version: pr5
+- src: ome.postgresql_backup
+  version: 0.2.0
 
 - src: ome.postgresql_client
   version: 0.1.2
--- a/requirements.yml
+++ b/requirements.yml
@@ -74,8 +74,9 @@
 - src: ome.postgresql
   version: 5.0.1
 
-- src: ome.postgresql_backup
-  version: 0.1.1
+- name: ome.postgresql_backup
+  src: https://github.com/ome/ansible-role-postgresql-backup/archive/acd03814b8b5dd891451407b28820bfbd4fadbc5.tar.gz
+  version: pr5
 
 - src: ome.postgresql_client
   version: 0.1.2
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -184,8 +184,7 @@
       tags: ssl
 
     - role: ome.postgresql_backup
-      postgresql_backup_dir: >-
-        {{ omero_server_db_dumpdir_parent | default('/tmp/pgbackup') }}
+      postgresql_backup_dir: /OMERO/pgbackup
       postgresql_backup_filename_format: "nightly-omero-%a.pgdump"
       postgresql_backup_minimum_expected_size: 1000000000
 
--- a/ome-demoserver.yml
+++ b/ome-demoserver.yml
@@ -183,6 +183,12 @@
     - role: ome.ssl_certificate
       tags: ssl
 
+    - role: ome.postgresql_backup
+      postgresql_backup_dir: >-
+        {{ omero_server_db_dumpdir_parent | default('/tmp/pgbackup') }}
+      postgresql_backup_filename_format: "nightly-omero-%a.pgdump"
+      postgresql_backup_minimum_expected_size: 1000000000
+
   handlers:
     - name: reload web server
       listen: ssl certificate changed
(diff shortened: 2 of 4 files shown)


sbesson:
Should `templates/nightly-pg_dump-omero.sh.j2` be dropped as well? Looking at

https://github.com/ome/prod-playbooks/blob/b04fb682f988dd0931d187f569bbe451661d6458/templates/nightly-pg_dump-omero.sh.j2#L4-L5

vs

https://github.com/ome/ansible-role-postgresql-backup/blob/3c2ccc78bd69b45f0cd4d5893e3e0a059a48d578/templates/etc-crond-postgresql-backup.j2#L4-L5

the major difference is the usage of `pg_dumpall` which unfortunately does not allow parallel dump. Do we foresee any issue associated with this?

manics:
I was going to remove it but it's also used on demo, though only to dump to `/tmp` https://github.com/ome/prod-playbooks/blob/b04fb682f988dd0931d187f569bbe451661d6458/ome-demoserver.yml#L378 so I thought I'd deal with nightshade first.

I'd expect the dump to take longer to run without any parallel options, but since it's non-interactive that shouldn't matter.

manics:
Note Travis is failing:
```
    RUNNING HANDLER [ome.omero_common : restart omero-server] **********************

    fatal: [ome-dundeeomero.openmicroscopy.org]: FAILED! => {"changed": false, "msg": "Unable to restart service omero-server: Job for omero-server.service failed because a timeout was exceeded. See \"systemctl status omero-server.service\" and \"journalctl -xe\" for details.\n"}
```

manics:
I think the failure was because the molecule test uses `/tmp` as the pg backup dir, but it changed the owner to `postgres` which caused OMERO to fail as it couldn't write to `/tmp`

sbesson:
:+1: for replacing custom tasks by an existing role. As a point of comparision in terms of performance, on `pub-omero`,

```
bash-4.2$ time pg_dump -j 4 -Fd -f /tmp/20200922_pg_dump_omero.dir/ omero

real	0m34.009s
user	1m6.756s
sys	0m1.188s
bash-4.2$ time pg_dumpall -f /tmp/20200922_pg_dumpall

real	0m34.115s
user	0m3.686s
sys	0m4.543s
```

For the functional testing, my proposal would be to add similar changes against `ome-demoserver.yml`, deploy them against `pub-omero` and test a database recovery scenario using the DB dump.

manics:
I've made the same changes to the demo playbook

sbesson:
With the systems back up and running, now deploy to `pub-omero.openmicroscopy.org` for internal testing. Proposed next steps:

- check that the daily DB dump is created in the expected place
- test a DB restore on `pub-omero`
- propagate the changes to the production systems `demo` + `ome-dundeeomero`

sbesson (changes requested):
As noted this morning, a side effect of https://github.com/ome/prod-playbooks/pull/285#issuecomment-709976471 has been to fill the `/tmp/` partition with daily 3G backups.

Next action will be to update the default backup dir to use the `/OMERO` partition or another mount which can accommodate 20-25G of dumps.

sbesson:
Deleted the previous backups on `/tmp/pgbackup/` of `pub-omero`. This inspired me a related thought: at the moment, the vanilla output of the backup role are uncompressed dumps. Is there an easy way to enable compression?

manics:
I've updated this PR to use https://github.com/ome/ansible-role-postgresql-backup/pull/5, `pub-omero` has been updated

sbesson:
Looks good after the first execution on `pub-omero`

```
[root@pub-omero sbesson]# du -csh /OMERO/pgbackup/*
404M	/OMERO/pgbackup/nightly-omero-Wed.pgdump.gz
404M	total
```

I'll leave the cron job to run ~5-7 days of rolling backups and then test a full restore early next week.

requirements.yml
```diff
@@ -74,8 +74,9 @@
 - src: ome.postgresql
   version: 5.0.1
 
-- src: ome.postgresql_backup
-  version: 0.1.1
+- name: ome.postgresql_backup
```
sbesson, line 77: This can be updated to `ome.postgresql_backup 0.2.0` which has been released with compression support.

sbesson (changes requested):
Once the requirements file is updated to use the new role release, proposing to schedule the deployment of this playbook on the various target servers

sbesson (approved):
Thanks @manics. Now deployed on both demo servers for confirmation tomorrow. I will deploy on `ome-dundeeomero` tomorrow evening if no issue is found with the dump logic

sbesson:
Deployed on `ome-dundeeomero.openmicroscopy.org`

PR #293, opened 2020-10-27
Add omedev-docker playbook

manics:
New docker playbook for Docker development VMs

Requires https://github.com/ome/ansible-role-docker/pull/8


sbesson:
The playbook looks fairly straightforward. [ome.docker 3.1.0](https://galaxy.ansible.com/ome/docker) has been released to Galaxy so the requirements file can be updated.

PR #296, opened 2020-12-07
Release of www on 7 Dec 2020

pwalczysko:
Deployed on 

  - ome-www-dev
  - ome-www

All seems to have worked fine, the deployment looks good.

cc @sbesson @joshmoore

### DIFF
--- a/www/www-static.yml
+++ b/www/www-static.yml
@@ -48,8 +48,8 @@
       path: "{{ phpbbforum_style_file }}"
 
   vars:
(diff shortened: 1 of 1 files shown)


sbesson:
Travis failure is unrelated. Merging to keep playbook in sync with the current deployment state. As discussed separately, we will likely be moving away from `ome.deploy_archive` imminently to manage the release frequency of www.openmicroscopy.org

PR #297, opened 2020-12-07
fixes typo in variable name -  https_letsencrypt_enabled

SteveO-sudo:
I noticed that the variable name at the end of the script has an underscore, while the actual variable that is checked in the script to conditionally run doesn't have an underscore. So, uncommenting the line at the end of the script and setting it to True doesn't actually do anything.


manics (approved):

manics:
Thanks for the fix! Travis failure is unrelated.

PR #300, opened 2021-01-07
ome-dockr-prod1.openmicroscopy.org prometheus grafana

manics:
## Runs Prometheus, Grafana in Docker.

This is a replacement for the monitoring applications in https://github.com/ome/kubernetes-apps/
Dashboards will need to be created/copied/edited manually

Requires:
- [ ] https://github.com/ome/ansible-role-nginx/pull/9
- [ ] https://github.com/ome/ansible-role-nginx-proxy/pull/32
- [ ] https://github.com/ome/ansible-role-prometheus/pull/9
- [ ] https://github.com/IDR/ansible-role-redmine-tracker/pull/1

## Runs minio in Docker

This is a replacement for https://github.com/ome/kubernetes-apps/


molecule/docker-prod/converge.yml
```diff
@@ -0,0 +1,19 @@
...
+- name: Converge
+  hosts: ome-dockr-prod1.openmicroscopy.org
+  tasks:
+    - name: Workaround to get host IP inside docker
+      shell: hostname -I | cut -d' ' -f1
+      register: hostname_ip
+      check_mode: false
+      changed_when: false
+      tags:
+        # Ignore [306] Shells that use pipes should set the pipefail option
```
joshmoore, line 11: But _shouldn't_ we be setting `pipefail`? ;)

joshmoore (commented):
Few questions:
 * How close are we to the limit of what this prod1 box can handle?
 * What's the relationship of this monitoring setup to that on Embassy?

molecule/docker-prod/converge.yml
manics: Could do, but I'd need to find out how, and in any case since the incorrect output would fail in the playbook

manics:
Total resource usage so far is fairly small:
image

The prometheus/grafana setup is fairly similar to EBI, the main difference is that hosts are enumerated individually instead of using Ansible host-groups as each host is slightly different

manics:
I've found a problem- the container IPs may change after a restart of Docker which breaks the Nginx proxy. There are two options:
1. Expose all ports externally- this is fine as long as no-one logs into redmine directly over that port since it's unencrypted
2. Run Nginx inside a Docker container on the same docker network as the other containers so only Nginx has an exposed port, but this means a custom Nginx config instead of using the standard proxy and HTTPS config from Ansible.

Any thoughts?

joshmoore:
If this docker were _solely_ for redmine, I could see 1. Guess we'd need to make sure everyone is aware of the repercussions if this is truly a "public dockr".

manics:
Nginx is now run in Docker. Be aware this means the logs are not saved to `/var/log/nginx`, however since this is an internal server that shouldn't be a concern.

joshmoore (approved):
I can't say I grok everything that's in the new playbook offhand, but I do think it looks _grokable_.

PR #303, opened 2021-01-27
Use idr.redmine_tracker from Ansible Galaxy

sbesson:
Following the publication of the role to https://galaxy.ansible.com/idr/redmine_tracker


sbesson:
Merging as GitHub Actions is green.

PR #324, opened 2021-11-10
Migrate demo deletion script to the public playbook

sbesson:
Motivated by https://forum.image.sc/t/how-to-clean-our-omero-data-repository/59836

This script is periodically used for cleaning up the OME demo server. The decision is based on a calculation of the maximal space consumption by users and might need to be adapted to various requirements but it contains a lot of logic for cleaning data that could be re-used and modified to suit various needs


joshmoore:
Two thoughts:
 * Are there enough caveats in place? (Guess the next level would be a banner or such)
 * Should this be a plugin in order to be versioned separately? (If so, I imagine we could keep "demo" in the name to add an extra level of caveat)

sbesson:
Yes turning this into a plugin (`omero demo-cleanup`?) would make sense in terms of packaging, versioning & deployment via `omero_server_python_addons` but also for others to consume & extend/fork. I primarily opened this PR to make the script available publicly but if we agree, I can start working on refactoring in the background and update/close this PR when ready.

In terms of caveats, I would assume the combination of an explicit plugin name + readme would be a start (for anyone running `pip install`). At execution time,  the default behavior is a dry-run until `--force` is passed so that's an extra level of safety that should prevent unexpected data deletion.

joshmoore:
Maybe we should get a cookiecutter in place which is the same thought I had _last time_ this came up.

sbesson:
The cleanup has now been migrated as a CLI plugin (https://github.com/ome/omero-demo-cleanup/actions, http://pypi.org/project/omero-demo-cleanup). I'll open another PR to install it into the relevant servers.

PR #327, opened 2021-11-23
Install newly released omero-demo-cleanup plugin into the demo server

sbesson:
Replaces https://github.com/ome/prod-playbooks/pull/324


sbesson:
Now deployed on `pub-omero` and `ome-demoserver`

PR #331, opened 2022-02-22
Add public samples for Leica-XLEF

dgault:
Adding the mapping for the new Leica XLEF public samples


sbesson (approved):
Deployed - see https://downloads.openmicroscopy.org/images/Leica-XLEF/. The CI failure is unrelated and this will require the last 2 commits from https://github.com/ome/prod-playbooks/pull/330#issuecomment-1048203708 to pass the  builds /cc @jburel

PR #344, opened 2022-03-28
bump omero-web version

jburel:
ns-web and ns-web-pub updated


sbesson (approved):
I understand this is now deployed to productio

PR #347, opened 2022-05-31
Add public user to web config

pwalczysko:
In exploration of adding of public user to outreach server via omero web config.
Does not work very well atm. When a new public user is created, and a new public group, the Dataset inside that group is not publicly visible, e.g. https://ome-training-4.openmicroscopy.org/webclient/?show=dataset-5405
Not sure where the mistake is.

cc @jburel @sbesson

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -77,7 +77,26 @@
 
     - role: ome.omero_web
       tags: ['web']
-      omero_web_config_set:
+        idr_omero_web_public_url_filters_webclient_exclude:
+          - action
+          - annotate_(file|tags|comment|rating|map)
+          - script_ui
+          - ome_tiff
+          - figure_script
+        idr_omero_web_public_url_filters:
+          - api/
+          - webadmin/myphoto/
+          - mapr/
+          - figure/
+          - iviewer/
+          - '$'
+          - gallery-api/
+          - gallery_settings/
+          - cell/
+          - tissue/
+          - webclient/(?!({{ idr_omero_web_public_url_filters_webclient_exclude | join('|') }}))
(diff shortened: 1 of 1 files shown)


sbesson:
Looking at a specific image in this dataset, https://ome-training-4.openmicroscopy.org/webgateway/render_image/53009/ works for me. Is the problem that `/webclient` is not in the `url_filter` list?

pwalczysko:
The playbook works now as expected, the url https://ome-training-4.openmicroscopy.org/webclient/?show=dataset-5405 is opening without creds in private window, thanks for the hint @sbesson . The situation is complicated by the fact that the full config is not present in our doc, see https://docs.openmicroscopy.org/omero/5.6.4/sysadmins/public.html#configuring-public-user contains only 

``omero config set omero.web.public.url_filter '^/(webadmin/myphoto/|webclient/(?!(script_ui|ome_tiff|figure_script))|webgateway/(?!(archived_files|download_as))|iviewer|api)'``

This is not matching the real situation on ns server, which has 

``omero.web.public.url_filter: ^\/(api|iviewer|figure|webadmin\/myphoto\/|webclient\/(?!(script_ui|ome_tiff|figure_script|render_image_download|download_|ome_tiff_script|archived_files\/download|download_orig_metadata))|webgateway\/(?!(archived_files|download_as)))``

And also, the latter (the ns situation) is not in this repo, but in the mngmt one.

sbesson (approved):
Looks good to me. LGTM assuming the deployment with the private variables was successful

pwalczysko:
Thank you @sbesson 

Successfully deployed and tested on all 4 outreach servers.
E.g. https://outreach.openmicroscopy.org/webclient/

Merging

PR #349, opened 2022-06-15
add logic for nightly pg dumps

pwalczysko:
cc @sbesson 

deployed on ome-training-3

possibly there will be some DB to use tomorrow ?

cc @khaledk2


pwalczysko:
@sbesson if the expectation of successful dump creation o/n is still reasonable ?, then I would like to do the suggested changes possibly later, as need to prioritize the upcoming workshop

PR #354, opened 2022-06-21
Disable PixelData

pwalczysko:
See https://github.com/ome/prod-playbooks/issues/353 for background.

@sbesson I am afraid the suggested change as implemented in this PR and tested on ome-training-4 is causing "No processor available" when trying to run a script (Batch image export).

### DIFF
--- a/omero/training-server/playbook.yml
+++ b/omero/training-server/playbook.yml
@@ -496,7 +496,7 @@
       omero.ldap.user_filter: "(objectClass=person)"
       omero.ldap.user_mapping: "omeName=uid,firstName=givenName,lastName=sn,email=mail"
       omero.ldap.username: "uid=admin,ou=system"
-      omero.server.nodedescriptors: "master:Blitz-0,Tables-0,Processor-0,Indexer-0"
+      omero.server.nodedescriptors: "master:Blitz-0,Indexer-0,Processor-0,Storm,Tables-0"
 
     external_nic: "{{ ansible_default_ipv4.interface }}"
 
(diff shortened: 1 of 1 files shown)


joshmoore:
ObjectAdapterDeactivatedException occurs during shutdown. Does the timestamp match a restart?

pwalczysko:
> ObjectAdapterDeactivatedException occurs during shutdown. Does the timestamp match a restart?

Thanks. Yes, it matches the time when I ran the playbook. So this is a red herring, i.e. not the cause of the problem ?

joshmoore:
> So this is a red herring, i.e. not the cause of the problem ?

I believe so.

pwalczysko:
Test: 
- rerunning the playbook without the changes in this PR restores the scripting ability on ome-training-4. Rerunning the playbook with changes in this PR again leads to repeated loss of the scripts to find the Processor
Screenshot 2022-06-28 at 15 13 26

Not sure where to go from here.

sbesson:
After some trial and error, it looks like `Storm` is a requirement for the `Processor` server to be functional. Below is the config that worked for me in the end.

```
omero.server.nodedescriptors master:Blitz-0,Indexer-0,Processor-0,Storm,Tables-0
```

I reverted my change and restarted the server, do you want to give it a try?

pwalczysko:
Thank you, will rerun tomorrow.

pwalczysko:
Applied the changes in the last commit to ome-training-4. Scripts, search of newly created KVPs and general webclient work fine. The Pixel data tests:
Imported https://ome-training-4.openmicroscopy.org/webclient/?show=image-56957 (no pyramids) - shows clock placeholder on thumb, no items with this ID in Pixels folder.
Imported https://ome-training-4.openmicroscopy.org/webclient/?show=image-56951 (svs with pyramids) - thumb and full viewer okay, no detriment in performance.

I think this is good, @sbesson ?

pwalczysko:
Deployed successfully on all 4 training servers. All seems to work. Thank you. Merging.

PR #358, opened 2022-07-08
training-server: extract database installation to sub-playbook

sbesson:


pwalczysko:
the molecule test failed twice already ( I have rerun the failing job once and it failed again).
Not sure what the error is

sbesson:
Forgot that the `postgresql_version` is also used by the server role to install postgres as a client and run PSQL commands. Closing this.

PR #360, opened 2022-07-08
Update training servers to PSQL 13

pwalczysko:
Upgrade postgres to 13.

Successfully deployed on ome-training-3

### DIFF
--- a/molecule/omero-training-server/tests/test_default.py
+++ b/molecule/omero-training-server/tests/test_default.py
@@ -13,7 +13,7 @@ OMERO_LOGIN = '-C -s localhost -u root -w omero'
     'nginx',
     'omero-server',
     'omero-web',
-    'postgresql-10',
+    'postgresql-13',
     'prometheus-node-exporter',
     'prometheus-omero-exporter',
     'prometheus-postgres-exporter',


sbesson (changes requested):
```
    TASK [ome.postgresql : postgres | install ansible prerequisites] ***************
    fatal: [ome-outreach]: FAILED! => {"changed": false, "msg": "No package matching 'python2-psycopg2' found available, installed or updated", "rc": 126, "results": ["No package matching 'python2-psycopg2' found available, installed or updated"]}
```

Can you also try bumping `ome.postgresql` to version `5.2.0` as this should have been fixed in https://github.com/ome/ansible-role-postgresql/pull/28

sbesson:
https://github.com/ome/prod-playbooks/blob/8c79952cbbcd3b03e7d16734dc1bc3ecf4d81ee7/molecule/omero-training-server/tests/test_default.py#L16 also needs to be updated to let the Molecule tests pass

sbesson:
Summarizing the sequence of actions that were performed on `ome-training-3` together with @pwalczysko:

1. stop the services: ` sudo service omero-web stop && sudo service omero-server stop && sudo service postgresql-10 stop`
2. execute this playbook including the PSQL version changes i.e. install the PSQL 13 packages and initializes a cluster. The playbook fails when starting the server as the DB is not initialized with the OMERO SQL schemas
3. as the `postgres` user, run `cd /OMERO/pgbackup && gunzip nightly-omero-Fri.pgdump.zip && psql -f nightly-omero-Fri.pgdump` to restore the DB under the PSQL 13 cluster
4. re-execute the Ansible playbook: the server and web services now start as expected
5. cleanup the old packages and cluster `sudo yum postgresql10* && sudo rm -rf /var/lib/psql/10`

jburel:
did those steps work when you replace pg13 by pg14?

sbesson:
Untested. As far as I am concerned the Ansible upgrade to PSQL14 is blocked by https://github.com/ome/ansible-role-postgresql/pull/31#issuecomment-1178893604.

The decision made together with @pwalczysko was to target PSQL13 as the most recent version compatible with the current role.

Once our fleet of PSQL 10 servers is upgraded to 13, we could certainly use one of the internal ones and test the upgrade to 14.

jburel:
understood

pwalczysko:
```
 deftest_service_running_and_enabled(host, name):
            service = host.service(name)
    >       assert service.is_running
E       assert False
E        +  where False = <service postgresql-10>.is_running
tests/test_default.py:23: AssertionError
    =========================== short test summary info ============================
    FAILED tests/test_default.py::test_service_running_and_enabled[ansible://ome-outreach-postgresql-10]
========================= 1 failed, 12 passed in 6.02s =========================
```

@sbesson I can see that the error here ^ is caused by still using the old psql version. It seems that I would know how to fix it, but then, how come that your PR https://github.com/ome/prod-playbooks/pull/361 with practically  the same commits passes the tests ? Maybe this is done by putting 13 into each variable for each playbook ? Maybe I should close this and work from the top of https://github.com/ome/prod-playbooks/pull/361 instead ?

sbesson (approved):

pwalczysko:
```
[pwalczysko@ome-training-4 lib]$ ls | grep pg
pgsql
[pwalczysko@ome-training-4 lib]$ sudo -u postgres -s
bash-4.2$ pwd
/var/lib
bash-4.2$ cd pgsql/
bash-4.2$ ls
10  13
```

@sbesson I think that the removal of the ``/var/lib/psql/10`` directory does not work as written (does nothing without any message I would claim).
See above for investigation (after I have run the ``rm -rf`` command as specified in https://github.com/ome/prod-playbooks/pull/360#issuecomment-1179182594 - I guess two things are missing:

1. psql -> pgsql
2. need to be postgres to be able to do anything ?

sbesson:
Yes, that's another typo, apologies. The folder to delete on RHEL7 is `/var/lib/pgsql/10` indeed.

The cluster deletion can be performed by the `postgres` user as it should own the directory. However the package removal needs root permissions.

pwalczysko:
Upgraded all 4 training servers according to https://github.com/ome/prod-playbooks/pull/360#issuecomment-1179182594

PR #361, opened 2022-07-11
Update learning/sls-gallery to PSQL 13

sbesson:
Includes some of the commits from #360 to bump the roles to versions supporting PSQL 13+.

Also migrates the `postgresql_version` under `vars` as it is transitively used in the `om.postgresql_client` dependency

### DIFF
--- a/omero/sls-gallery.yml
+++ b/omero/sls-gallery.yml
@@ -26,6 +26,12 @@
           password: "{{ omero_server_dbpassword | default('omero') }}"
           databases: [omero]
 
+    - role: ome.postgresql_backup
+      postgresql_backup_compress: true
+      postgresql_backup_dir: /OMERO/pgbackup
+      postgresql_backup_filename_format: "nightly-omero-%a.pgdump.gz"
+      postgresql_backup_minimum_expected_size: 100000000
+
     - role: ome.lvm_partition
       lvm_vgname: VolGroup00
       lvm_lvname: 'omero'


sbesson:
On `ome-learning`, given I had less than 50% space on the `/var/lib/pgsql` partition, I had to go for a modified version of https://github.com/ome/prod-playbooks/pull/360#issuecomment-1179182594

- have a VM snapshot and check the nightly PG dump is presen int
- stop the services `sudo service omero-web stop && sudo service omero-server stop && sudo service postgresql-10 stop`
- run the playbook until the completion of the `ome.postgresql` role to install the new packages/initialize the cluster
- as the `postgres` user, first delete the existing cluster `rm -rf /var/lib/pgsql/10` then unzip and restore the dump `cd /OMERO/pgbackup && gunzip nightly-omero-Mon.pgdump.gz && psql -f nightly-omero-Mon.pgdump && gzip nightly-omero-Mon.pgdump`
- run the playbook again until completion to restart all services
- cleanup the former packages `sudo yum remove postgresql10*`

joshmoore:
It sounds a _little_ bit like the openstack creation routines which can choose what their source "volume" is. Here it's source "database" which could be local if there's enough space but otherwise is remote and needs to be copied/unzipped into place.

sbesson:
> It sounds a little bit like the openstack creation routines which can choose what their source "volume" is. Here it's source "database" which could be local if there's enough space but otherwise is remote and needs to be copied/unzipped into place.

In the back of my mind, I was considering how to perform such an upgrade in the case of IDR where we are at also over 50% usage

```
[sbesson@prod107-database ~]$ df -h | grep pgsql
/dev/vdb        800G  553G  248G  70% /var/lib/pgsql
```

A source volume containing a dump of the database might effectively mirror what happens here.

PR #362, opened 2022-08-11
Bump omero_figure_release to 5.0.0 everywhere

will-moore:
Bump OMERO.figure version to latest 5.0.0 everywhere.

cc @pwalczysko

### DIFF
--- a/omero/nightshade-webclients.yml
+++ b/omero/nightshade-webclients.yml
@@ -103,7 +103,7 @@
       }}
 
     omero_web_release: "{{ omero_web_release_override | default('5.14.1') }}"
-    omero_figure_release: "{{ omero_figure_release_override | default('4.4.3') }}"
+    omero_figure_release: "{{ omero_figure_release_override | default('5.0.0') }}"
     omero_fpbioimage_release: "{{ omero_fpbioimage_release_override | default('0.4.0') }}"
     omero_iviewer_release: "{{ omero_iviewer_release_override | default('0.11.3') }}"
     omero_parade_release: "{{ omero_parade_release_override | default('0.2.3') }}"
--- a/omero/ome-demoserver.yml
+++ b/omero/ome-demoserver.yml
@@ -310,7 +310,7 @@
 
   vars:
     omero_figure_release: "{{ omero_figure_release_override | default('5.0.0') }}"
-    omero_figure_script_release: "{{ omero_figure_script_release_override | default('v4.4.3') }}"
+    omero_figure_script_release: "{{ omero_figure_script_release_override | default('v5.0.0') }}"
     omero_fpbioimage_release: "{{ omero_fpbioimage_release_override | default('0.4.0') }}"
     omero_iviewer_release: "{{ omero_iviewer_release_override | default('0.11.3') }}"
     omero_parade_release: "{{ omero_parade_release_override | default('0.2.3') }}"
--- a/omero/ome-demoserver.yml
+++ b/omero/ome-demoserver.yml
@@ -309,7 +309,7 @@
         force: yes
 
   vars:
-    omero_figure_release: "{{ omero_figure_release_override | default('4.4.3') }}"
+    omero_figure_release: "{{ omero_figure_release_override | default('5.0.0') }}"
     omero_figure_script_release: "{{ omero_figure_script_release_override | default('v4.4.3') }}"
     omero_fpbioimage_release: "{{ omero_fpbioimage_release_override | default('0.4.0') }}"
     omero_iviewer_release: "{{ omero_iviewer_release_override | default('0.11.3') }}"
--- a/omero/ome-dundeeomero.yml
+++ b/omero/ome-dundeeomero.yml
@@ -208,7 +208,7 @@
     nginx_version: 1.18.0
     postgresql_version: "11"
     filesystem: "xfs"
-    omero_figure_release: "{{ omero_figure_release_override | default('4.4.3') }}"
+    omero_figure_release: "{{ omero_figure_release_override | default('5.0.0') }}"
 
     omero_server_config_set_production:
       omero.db.poolsize: 60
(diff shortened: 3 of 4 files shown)


sbesson:
:+1: Similar question as always with OMERO.figure: does the figure script need to be updated i.e. do we need a server restart in addition to the OMERO.web restart?

will-moore:
Yes, the script does need updating. But I'm not familiar with how that happens on our playbooks (normally a script update wouldn't require a server restart)

sbesson (approved):
The script updates are standalone tasks similar to https://github.com/ome/prod-playbooks/blob/83513f1010cdf648cb5fe4c9a09ed1ed9a0eedb0/omero/ome-dundeeomero.yml#L177-L184 which are consuming the `omero_figure_release` variable that you modified already. So no additional change is required.

Looking at the tasks, you are correct a server restart is not required when this task is updated. This is useful and allows us to apply these upgrades without downtime

@pwalczysko any preference on how to proceed here for the deployment?

pwalczysko:
> @pwalczysko any preference on how to proceed here for the deployment?

I know how to do the training servers, as the playbook is just one there for omero and web.
But for ns, my experience is limited to server upgrade deployment:
https://github.com/ome/prod-playbooks/pull/357

Is it feasible to run just the web-related playbooks on ns ? (which ones exactly if yes ?)
About the demo, never did it yet.

sbesson:
> Is it feasible to run just the web-related playbooks on ns ? (which ones exactly if yes ?)
About the demo, never did it yet.

Practically, it's the same process as the server except you are running other playbooks with different scopes:

- `nightshade-webclients.yml`  manages the UoD SLS OMERO.web deployments (authenticated and public)
- `ome-demoserver.yml` manages the OME demo OMERO.server and its OMERO.web deployment

I expect you have full permissions to be able to run both playbooks. That being said, it would be very timely to confirm this is the case and fix issues if necessary.

pwalczysko:
@sbesson Ran the two playbooks (from this branch) you mention in https://github.com/ome/prod-playbooks/pull/362#issuecomment-1211899615 against three servers in check mode:

```
ns-web.openmicroscopy.org
ns-web-pub.openmicroscopy.org
ome-demoserver.openmicroscopy.org
```

The output looked reasonable and there were no probs with authentication

pwalczysko:
I will hold on upgrading further servers until the problem with ``Figure_To_Pdf.py`` is clarified.
At the moment, we have 

```
Traceback (most recent call last):
  File "/opt/omero/web/venv3/lib64/python3.6/site-packages/omeroweb/webclient/views.py", line 4963, in run_script
    handle = scriptService.runScript(sId, inputMap, None, conn.SERVICE_OPTS)
  File "/opt/omero/web/venv3/lib64/python3.6/site-packages/omero/gateway/__init__.py", line 4796, in __call__
    return self.handle_exception(e, *args, **kwargs)
  File "/opt/omero/web/venv3/lib64/python3.6/site-packages/omeroweb/webclient/webclient_gateway.py", line 2109, in handle_exception
    super(OmeroWebSafeCallWrapper, self).handle_exception(e, *args, **kwargs)
  File "/opt/omero/web/venv3/lib64/python3.6/site-packages/omero/gateway/__init__.py", line 4793, in __call__
    return self.f(*args, **kwargs)
  File "/opt/omero/web/venv3/lib64/python3.6/site-packages/omero_api_IScript_ice.py", line 935, in runScript
    return _M_omero.api.IScript._op_runScript.invoke(self, ((scriptID, inputs, waitSecs), _ctx))
omero.InternalException: exception ::omero::InternalException
{
    serverStackTrace = 
    serverExceptionClass = 
    message = Sha1s don't match! expected 0d195d208b1493d8997c4d04c68813936eb81db2, found 15f37a30ba393bc054aad734154c99cfedc58425
}
```

When the attempt is made to export the pdf from the OMERO.figure UI on ome-training-4.

sbesson:
Thanks @pwalczysko, it looks like the replacement of the script file directly within the server directory is not sufficient and additional logic is required to update the internal checksums. Possibly previous OMERO.figure upgrades were combined with OMERO.server upgrades so that the service restart was triggered anyways and we did not encounter this error.

Options I can think of:
- add a server restart handler to the script replacement task
- invoke an OMERO CLI command (`omero script upload/replace/edit`) for replacing the script content
- find an intermediate command that would refresh the script list without requiring a full server restart

will-moore:
As discussed this morning (@sbesson  and @pwalczysko), plan for nightshade is to upgrade omero-web via the playbook. At the same time, update the figure script manually via `$ omero script replace ...` which won't require any server downtime.

will-moore:
`omero script replace...` failed on nightshade with
```
  File "/Users/wmoore/opt/anaconda3/envs/omeroweb/lib/python3.9/site-packages/omero_api_IScript_ice.py", line 767, in editScript
    return _M_omero.api.IScript._op_editScript.invoke(self, ((fileObject, scriptText), _ctx))
omero.InternalException: exception ::omero::InternalException
{
    serverStackTrace = java.io.IOException: File '/opt/omero/server/OMERO.server-5.6.5-ice36-b233/lib/scripts/omero/figure_scripts/Figure_To_Pdf.py' cannot be written to
	at org.apache.commons.io.FileUtils.openOutputStream(FileUtils.java:2177)
```
Seb: might be due the fact `/opt/omero/server` is owned by `root`

So we resorted to replace script via playbook and a restart of the server.

pwalczysko:
For the record, the 

> At the same time, update the figure script manually via $ omero script replace ... which won't require any server downtime.

did not work because of the folder permissions. The route of running the ``omero/ome-dundeeomero.yml -l ome-dundeeomero.openmicroscopy.org`` was taken, which elicited the Sha1 problem, and then the server was restarted.


Atm, all the servers (4 outreach servers, demo, ns-private and ns-public) are upgraded and working as expected.

pwalczysko (approved):

PR #364, opened 2022-09-07
Omeropy 5.12.0 web 5.15

will-moore:
Updating recent releases of omero-web and omero-py (which needs to use `omero_web_python_addons`)

cc @sbesson @pwalczysko

### DIFF
--- a/omero/learning.yml
+++ b/omero/learning.yml
@@ -169,7 +169,7 @@
   vars:
     postgresql_version: "13"
     omero_server_release: 5.6.5
-    omero_web_release: 5.14.1
+    omero_web_release: 5.15.0
     omero_web_apps_release:
       omero_gallery: 3.4.2
       omero_iviewer: 0.11.3
--- a/omero/learning.yml
+++ b/omero/learning.yml
@@ -115,6 +115,7 @@
         - omero-virtual-microscope=={{ omero_web_apps_release.omero_virtual_microscope }}
       omero_web_python_addons:
         - "django-redis==5.0.0"
+        - "omero-py>={{ omero_py_release }}"
 
     - role: ome.postgresql_backup
       postgresql_backup_compress: true
@@ -170,6 +171,7 @@
     postgresql_version: "13"
     omero_server_release: 5.6.5
     omero_web_release: 5.15.0
+    omero_py_release: "{{ omero_py_release_override | default('5.12.0') }}"
     omero_web_apps_release:
       omero_gallery: 3.4.2
       omero_iviewer: 0.11.3
--- a/omero/nightshade-webclients.yml
+++ b/omero/nightshade-webclients.yml
@@ -21,6 +21,8 @@
     # OMERO.web configuration in host_vars in different repository
     - role: ome.omero_web
       omero_web_systemd_limit_nofile: 16384
+      omero_web_python_addons:
+        - "omero-py>={{ omero_py_release }}"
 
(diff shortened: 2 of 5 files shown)


pwalczysko:
the https://github.com/ome/prod-playbooks/pull/364/commits/1cca67b5144df3f28efc3e09f1361f27a62e07e2 has a confusing name. It should be ``bump ... 5.15.0 on all servers``

sbesson (approved):
The proposed changes should be the canonical way to use the public role variables to set `omero-py` version in OMERO.server/OMERO.web environments. Looks good from a code review perspective and obviously should be confirmed by an execution of the playbook in a test environment.

pwalczysko:
@will-moore could you please merge https://github.com/will-moore/prod-playbooks/pull/2 ?

The training servers are not upgraded for web 5.15.x

pwalczysko:
```
ERROR]: failed to download the file: The read operation timed out
[WARNING]: - ome.certbot was NOT installed successfully.
ERROR! - you can use --ignore-errors to skip failed roles and finish processing the list.
Retrying dependency ... 2/3 time(s)
```

@sbesson ^ is this the correct error please ?

sbesson:
That's an `ansible-galaxy install` error, right? In which case restarting the command should suffice.

pwalczysko:
On the omero-server venv though, I have still

```
(venv3) [pwalczysko@ome-training-3 ~]$ pip freeze | grep omero-py
omero-py==5.10.1
```

I have checked also imports and insight and CLI.
Do we want to upgrade the omero-py on the server too ?

will-moore:
@pwalczysko I don't know that we need to update on the server. Hopefully be restricting this to web env we don't have to restart the server (no nightshade downtime etc)?
However, if it keeps things simpler for the future (less confusing etc) then I'm happy to bump it there too

pwalczysko:
> @pwalczysko I don't know that we need to update on the server. Hopefully be restricting this to web env we don't have to restart the server (no nightshade downtime etc)? However, if it keeps things simpler for the future (less confusing etc) then I'm happy to bump it there too

I would argue that not having the ``omero-py`` on the latest released version on the server is not what we usually do ?
"no ns downtime" does not seem like a strong reason to me. 
What is your opinion @sbesson please ?

sbesson:
> I would argue that not having the omero-py on the latest released version on the server is not what we usually do ?
"no ns downtime" does not seem like a strong reason to me.
What is your opinion @sbesson please ?

I think our policy differs depending on the  deployment strategy:

- for IDR, we use a typical blue/green deployment, so the next production deployment is typically provisioned with the current OMERO.py version. Given the monthly release frequency, we ensure OMERO.py is typically up-to-date
- for the UoD deployments, we use more classic strategy involving downtime & upgrades as necessary. As the base OMERO.server/OMERO.web role do not upgrade OMERO.py if it is already installed in the environment, at the moment, this means that the OMERO.py version on these systems will not track the latest  unless we explicitly upgraded either via playbook as proposed here or manually

pwalczysko:
Okay, so I guess the consensus is to leave the PR as is.

I have upgraded all the training servers now using this PR checkout. All looks good.

pwalczysko:
So here is the proposed plan

1. [ ] training-servers (4x)
2. [ ] demo
3. [ ] nightshade webclients (2x)
4. [ ] send email to learning

pwalczysko:
All is upgraded now successfully afaik.

pwalczysko (approved):

PR #372, opened 2023-07-25
Upgrade demo cleanup script

pwalczysko:
Upgrading the cleanup script on the demo server.
Now, only the ``. /opt/omero/server/venv3/bin/activate`` is necessary to be able to run the script.

This PR is necessary as the venv established in the home folders of @dominikl @pwalczysko and @khaledk2 were taking too much space on the root partition of the server.

This was run successfully on demo server.

cc @sbesson @jburel


sbesson (approved):
From a playbook perspective, the change seems straightforward. I am no longer managing the OME demo server and/or tracking the updates to the demo cleanup PR so I cannot comment on the feature itself. Maybe these should be better reviewed by  @ome/dundee members ?

PR #376, opened 2024-08-23
Fix README

pwalczysko:
Fixes https://github.com/ome/prod-playbooks/issues/375 - the README is outdated in 2 points:

1. It does not specify that the playbooks are working with CentOS 7 OS.
2. It points to ``ome-ansible-molecule`` which is not used anymore

@MartinDerHettich

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 OME production services playbooks
 =================================
 
+**NOTE: We are in the process of migration of the playbooks in this repo to Rocky Linux 9/RHEL 9 OS but at the moment, they function only on CentOS 7 !!**
 These playbooks encapsulate the running of various production servers run by the OME team.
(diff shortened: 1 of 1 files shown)


pwalczysko:
@sbesson imho, although the tests are failing (I believe because the CentOS 7 mirror cannot be found as CentOS is EOF), this PR should prevent needless install attempts such as https://github.com/ome/prod-playbooks/issues/375

joshmoore:
I'm going to merge as an immediate improvement on the former situation. There are definitely a number of additional options we can then consider:

- update the playbooks with the in-progress RHEL9 versions
- separate the playbooks into separate versioned directories
- archive the repo with a note pointing to the new starting point (perhaps just the [roles](https://github.com/ome/ansible-role-omero-server) [themselves](https://github.com/ome/ansible-role-omero-web))

PR #377, opened 2024-10-23
Update qa2

jburel:
Changes were manually made in ```proxy-default.conf``` on ome-www
Making changes to playbook so they do not get missed during the migration

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 roles
 .vagrant
+.DS_Store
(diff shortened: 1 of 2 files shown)


pwalczysko:
I think we migrated this to private repo, is that true @jburel ?

pwalczysko:
> I think we migrated this to private repo, is that true @jburel ?

I checked and I think the answer is NO.
Could we migrate this to private repo ? 
I would be in favor of this solution, as I am not sure the QA is of much public interest.

pwalczysko:
Actually I think that this PR is superseded by https://github.com/openmicroscopy/management_tools/pull/1768

PR #379, opened 2024-11-15
Reduce the prod-playbooks

pwalczysko:
This PR is replacing https://github.com/ome/prod-playbooks/pull/378

The idea is to slim down the prod-playbooks repo to a more manageable size of 1-2 playbooks which are offered to community as examples. 

The plan (to be executed in 2 stages) is to 
1. adjust and add the ``ome-demoserver.yml`` playbook and test it via molecule
2. adjust and add the training server playbook and test it via molecule

This PR is tackling step 1. above:

- Delete all the playbooks except the ``ome-demoserver.yml``
- Update the ``ome-demoserver.yml`` playbook for Rocky Linux 9 / RHEL 9 as it is used now in our private repo in production
- Add appropriate requirements.yml file
- Drop the ``site.yml`` logic as it is too complex for just 1-2 playbooks and does not help the users when attempting to run/use this repo imho
- Unify the ``templates`` with the private repo ones -> this involves update of ``omero-web-config-for-webapps.j2`` (tagsearch and webtagging)
- Add full test sequence (including ``create``, ``converge``..., i.e. the install of OMERO.server and OMERO.web as per ``ome-demoserver.yml`` playbook) to the ``molecule/ome-demoserver/molecule.yml``
- Adjust the README to reflect the new purpose and scope of this repo
- Make the changes neccessary for the molecule test above to pass, including
        - create a ``prepare.yml`` for install of ``cron``
        - add the ``inventory`` paragraph to ``molecule.yml``to pass the correct variable values to the testing framework
        - make sure the roles during the test are installed where the molecule machinery can find them (change in ``ansible.cfg``)
        - run the dependency step before the first destroy step, as otherwise the destroy step crashes

I have also verified that the playbook ``ome-demoserver.yml`` is working fine on the production demo server in check mode when run from this repo exactly as prescribed now in the README cmds.

cc @khaledk2 @sbesson @jburel 
        -


requirements.yml
```diff
@@ -1,107 +1,24 @@
...
-- name: ome.omero_server
-  version: 4.0.2
-
-- src: ome.omero_user
-  version: 0.3.0
-
-- name: ome.omero_web
-  version: 4.0.1
-
+- src: ome.java
```
sbesson, line 4: Are these expected to be examples or the actual playbooks used for deploying the production services?
In the latter case, I think you might want to keep the version rather than installing the latest role by default

README.md
```diff
@@ -1,8 +1,12 @@
 OME production services playbooks
 =================================
 
-**NOTE: We are in the process of migration of the playbooks in this repo to Rocky Linux 9/RHEL 9 OS but at the moment, they function only on CentOS 7 !!**
 These playbooks encapsulate the running of various production servers run by the OME team.
+
+At the moment, only the `ome-demoserver.yml` playbook is available here. This is a playbook for OMERO.demo server on https://demo.openmicroscopy.org OMERO.server and OMERO.web. You can read more about this [OMERO.demo server on our website](https://www.openmicroscopy.org/explore/).
+
+We are in the process of adding more OME team's production playbooks here.
```
sbesson, line 8: Is the intent of this repository to only hold OMERO production playbooks moving forward?

pwalczysko: > Is the intent of this repository to only hold OMERO production playbooks moving forward?

That was my initial intent when planning this PR and discussing the tactics previously with @jburel .
What else would you like to hold here @sbesson ?
The idea is to ``minimize and simplify`` wherever possible.
But happy to discuss about better strategies.

requirements.yml
pwalczysko: Pin the versions of roles:

The drawback of this is that the versions will become stale (that happened already).
The advantage of this - I am not sure, maybe we should discuss this ? I thought our roles are pretty well maintained and we do not want to introduce anything to them which is not compliant with our prod playbooks deployment ?

README.md
sbesson: I don't have anything different to propose. I was mostly enquiring as I saw the `omero` subfolder was maintained and this was historically used for separating the OMERO playbooks from the others.

pwalczysko: Yes, ``omero`` folder is maintained to get out of the conandrum of molecule not finding the roles folder. Maybe I could rename the folder to ``playbooks`` ?

requirements.yml
sbesson: > The drawback of this is that the versions will become stale (that happened already).

That's correct, if these are stored, these versions need to be maintained, that's not different from any other dependency.

>  The advantage of this - I am not sure, maybe we should discuss this ?

Primarily a reproducibility argument. If a new version of the role with backwards-incompatible changes is made, this approach means installing the roles and running the playbook from the exact same commit would lead to different outcome.

dominikl:
👍 For the demo server email changes!

requirements.yml
pwalczysko: fixed in https://github.com/ome/prod-playbooks/pull/379/commits/25533d4a3fc1fe6f4ab84a177e6875253addef50

pwalczysko:
The fixes were pushed, the new changes are implemented too.

.github/workflows/molecule.yml
```diff
@@ -8,34 +8,33 @@ on:
...
+    runs-on: ubuntu-22.04
     strategy:
       # Keep running so we can see if other tests pass
       fail-fast: false
       matrix:
         scenario: ${{fromJson(needs.list-scenarios.outputs.matrix)}}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: actions/setup-python@v4
```
jburel, line 31: ``actions/setup-python`` bump to v5

pwalczysko: fixed in https://github.com/ome/prod-playbooks/pull/379/commits/fee1fdea5cbc1687f0ad82ef8205ba96d1d223ad

PR #390, opened 2026-02-19
Molecule specific tasks

pwalczysko:
This PR:

- bumps the ome.omero_server version to 7.0.1 as it is crucial for functioning to have ``omego==0.8.0``
- introduces the variable ``RUNNING_IN_MOLECULE`` to discern the molecule-specific tasks and uses it for the ``omero_server_selfsigned_certificates`` values (this helps with the usage of the playbook for our prod server, where the selfsigned is always ``false``, whereas in molecule it must be true).
- adds installation of ``dirmngr`` for ubuntu, as this is necessary to run the java installation without the previous ome.postgresql role usage in the playbook
- bumps the versions of server, web, omero-py and webapps

cc @khaledk2

### DIFF
--- a/playbooks/ome-demoserver.yml
+++ b/playbooks/ome-demoserver.yml
@@ -233,7 +233,8 @@
   
   vars:
     # python_version: "3.12"
-    omero_server_selfsigned_certificates: true
+    omero_server_selfsigned_certificates: >-
+      {{ (lookup('env', 'RUNNING_IN_MOLECULE') == "true") | default(false) }}
     icessl_default_dir_self: "{{ omero_server_basedir }}/selfsigned"
     icessl_default_dir_noself: "/etc/ssl/omero"
     icessl_default_dir: "{{ icessl_default_dir_self if omero_server_selfsigned_certificates else icessl_default_dir_noself }}"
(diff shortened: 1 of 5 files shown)


playbooks/ome-demoserver.yml
```diff
@@ -233,7 +233,8 @@
   
   vars:
     # python_version: "3.12"
-    omero_server_selfsigned_certificates: true
+    omero_server_selfsigned_certificates: >-
+      {{ (lookup('env', 'RUNNING_IN_MOLECULE') == "true") | default(false) }}
     icessl_default_dir_self: "{{ omero_server_basedir }}/selfsigned"
     icessl_default_dir_noself: "/etc/ssl/omero"
     icessl_default_dir: "{{ icessl_default_dir_self if omero_server_selfsigned_certificates else icessl_default_dir_noself }}"
```
khaledk2, line 250: Could you please also add a description for this variable?

pwalczysko: fixed in https://github.com/ome/prod-playbooks/pull/390/commits/17f00b6f6ef0e4ac933e4f9f6ef30404567138bd

khaledk2 (approved):

